PR #8, opened 2024-12-25
Create MarkerCommands

rakrakon:


src/main/kotlin/frc/robot/lib/MarkerCommands.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import edu.wpi.first.wpilibj.Alert
+import edu.wpi.first.wpilibj.DataLogManager
+import edu.wpi.first.wpilibj2.command.Command
+import edu.wpi.first.wpilibj2.command.Commands
+
+const val ABNORMAL_EVENT_NAME = "PROBLEM!"
+val alert = Alert(ABNORMAL_EVENT_NAME, Alert.AlertType.kWarning)
+
+fun markEvent(eventName: String): Command = Commands.runOnce({
```
katzuv, line 11: Can top-level functions in Kotlin be private? If yes, then this one should be private.

src/main/kotlin/frc/robot/lib/MarkerCommands.kt
```diff
@@ -0,0 +1,18 @@
...
+    DataLogManager.log(eventName)
+    alert.set(true)
+}).withTimeout(5.0).andThen(Commands.runOnce({
+    alert.set(false)
```
katzuv, line 15: Is this just for logging purposes? If so, will an alert be shown to the drivers?

rakrakon: This is for logging purposes. The driver calls this command when there's issues and it'll save the timestamp.

src/main/kotlin/frc/robot/lib/MarkerCommands.kt
rakrakon: 920e64150027ab79e600bc7fcca4bd1bc8aef948

src/main/kotlin/frc/robot/lib/MarkerCommands.kt
katzuv: I wonder if it might distract the driver. Because we do want to use the alerts minimally so the drivers will actually look at them when critical errors happen.

rakrakon: If we open the timestamp in advantagescope we can see when it exists and when it doesn't

katzuv: Okay. Still I'm not sure why you need this as an alert? Why is using `log()` not sufficient?
If the problem is that multiple messages will overlap, you can just add the timestamp or an index number to the message.

rakrakon: When we do it as alert we can make it show up in advantagescope. If we log it we'll have to search through the logs for when it happend.

katzuv: 1. You still log it
2. The messages log is used by team code only, it shouldn't have other types of messages.
3. Errors and warnings highlighted by AdvantageScope.
3. You can use another message entry for the driver markers.

I'm aware of the alert feature 🙂 but this is not the use case here, they should only be used to notify the drivers.

katzuv (approved):

src/main/kotlin/frc/robot/lib/MarkerCommands.kt
```diff
@@ -0,0 +1,22 @@
...
+
+const val ABNORMAL_EVENT_NAME = "PROBLEM!"
+
+@AutoLogOutput
+private var marked = false
+
+private fun markEvent(eventName: String): Command = Commands.runOnce({
+    DataLogManager.log(eventName)
+    marked = true
+}).withTimeout(5.0).andThen(Commands.runOnce({
```
katzuv, line 18: I think you might need to use the `waitSeconds()`instead because `runOnce()` ends immediately.

katzuv (changes requested):

src/main/kotlin/frc/robot/lib/MarkerCommands.kt
rakrakon: ddab531ec3d111387673215a769afe8fbc1912cc

rakrakon: I'll test it later

katzuv: Ok, so please merge only after testing.

src/main/kotlin/frc/robot/lib/MarkerCommands.kt
```diff
@@ -0,0 +1,22 @@
+package frc.robot.lib
+
+import edu.wpi.first.wpilibj.Alert
+import edu.wpi.first.wpilibj.DataLogManager
+import edu.wpi.first.wpilibj2.command.Command
+import edu.wpi.first.wpilibj2.command.Commands
+import org.littletonrobotics.junction.AutoLogOutput
+import org.littletonrobotics.junction.Logger
+
+const val ABNORMAL_EVENT_NAME = "PROBLEM!"
```
katzuv, line 10:
```suggestion
const val ABNORMAL_EVENT_NAME = "### [Error] DRIVER PROBLEM REPORTED ###"
```
Reasons for the suggestions:
1. It's important to include "error"/"warning" in the message so AdvantageScope will syntax-highlight it:
   
2. The message should be a bit more descriptive and visible.

katzuv (changes requested):

katzuv (approved):

PR #55, opened 2025-01-15
feature/elevator

DavidMasin:


Emma03L:
You should change the base to dev and rename this pr to something more indicative

src/main/kotlin/frc/robot/subsystems/elevator/ElevatorIOReal.kt
```diff
@@ -0,0 +1,61 @@
...
+    override fun setPower(percentOutput: Double) {
+        motor.set(percentOutput)
+    }
+
+    override fun reset() {
+        motor.setPosition(0.0)
+    }
+    override fun updateInputs() {
+        inputs.appliedVoltege = motor.motorVoltage.value
+        inputs.height = Units.Centimeter.of(motor.position.value.`in`(Units.Rotations) * ROTATIONS_TO_CENTIMETER)
```
katzuv, line 59: Can you find a way to make this shorter? Perhaps using [`Per`](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/units/measure/Per.html). @rakrakon @emma03l your help will be appreciated here too, I didn't find a way to make it work. Basically we want a way to convert an angle measure (received from the motor) to a distance measure (height) without creating a new measure every loop.
At the very least, use a mutable measure. But I think with some `Per` magic and using `timesDivisor()` and/or `timesConversionFactor()` we can solve this. Thanks!

katzuv (changes requested):

katzuv:
Closing in favor of #61.

PR #67, opened 2025-01-16
Measure extension functions

Emma03L:
Add extension functions that convert Distance measures to Angle and vice versa

### DIFF
--- a/src/main/kotlin/frc/robot/lib/ExtensionFunctions.kt
+++ b/src/main/kotlin/frc/robot/lib/ExtensionFunctions.kt
@@ -13,9 +13,9 @@ import edu.wpi.first.units.measure.Distance
 import edu.wpi.first.wpilibj2.command.Command
 import edu.wpi.first.wpilibj2.command.WrapperCommand
 import frc.robot.IS_RED
+import org.littletonrobotics.junction.LogTable
 import kotlin.math.PI
 import kotlin.math.hypot
-import org.littletonrobotics.junction.LogTable
 
 fun ChassisSpeeds.getSpeed() = hypot(vxMetersPerSecond, vyMetersPerSecond)
 
@@ -107,6 +107,4 @@ fun Distance.toAngle(radius: Distance): Angle =
     )
 
 fun Angle.toDistance(radius: Distance): Distance =
-    Units.Meters.of(
-        this.`in`(Units.Rotations) * 2 * PI * radius.`in`(Units.Meters)
-    )
+    radius.times(this.`in`(Units.Rotations) * 2 * PI)
--- a/src/main/kotlin/frc/robot/lib/ExtensionFunctions.kt
+++ b/src/main/kotlin/frc/robot/lib/ExtensionFunctions.kt
@@ -102,7 +102,7 @@ fun Rotation2d.flip(): Rotation2d = FlippingUtil.flipFieldRotation(this)
 fun Rotation2d.flipIfNeeded(): Rotation2d = if (IS_RED) this.flip() else this
 
 fun Distance.toAngle(radius: Distance): Angle =
-    Units.Degree.of(
+    Units.Rotations.of(
         this.`in`(Units.Meters) / (2 * PI * radius.`in`(Units.Meters))
     )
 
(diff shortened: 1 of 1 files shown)


src/main/kotlin/frc/robot/lib/ExtensionFunctions.kt
```diff
@@ -96,3 +100,13 @@ fun Translation2d.flipIfNeeded(): Translation2d =
 fun Rotation2d.flip(): Rotation2d = FlippingUtil.flipFieldRotation(this)
 
 fun Rotation2d.flipIfNeeded(): Rotation2d = if (IS_RED) this.flip() else this
+
+fun Distance.toAngle(radius: Distance): Angle =
+    Units.Degree.of(
+        this.`in`(Units.Meters) / (2 * PI * radius.`in`(Units.Meters))
```
katzuv, line 106: You can divide meters by meters, you don't need to convert to doubles. However I think you need to do `Units.Rotations()`. Think what happens if you do `Meters.of(2πr, r).toAngle()`. You should get an angle measure of one rotation.

src/main/kotlin/frc/robot/lib/ExtensionFunctions.kt
```diff
@@ -96,3 +100,13 @@ fun Translation2d.flipIfNeeded(): Translation2d =
...
+    )
+
+fun Angle.toDistance(radius: Distance): Distance =
+    Units.Meters.of(
+        this.`in`(Units.Rotations) * 2 * PI * radius.`in`(Units.Meters)
```
katzuv, line 111: Just use `in.radians()` if that's what you want.

katzuv (changes requested):

src/main/kotlin/frc/robot/lib/ExtensionFunctions.kt
Emma03L: I think it's more readable like this because multiplying rotations by 2 pi r is more intuitive than multiplying radians by r

src/main/kotlin/frc/robot/lib/ExtensionFunctions.kt
Emma03L: `Rotation.of()` only accepts a double and I need to multiply by 2 pi

Emma03L: But I did fix the units 2b4cfdd286f19d1369f72f6fa8c93b00f7b18d89

katzuv: Can't you do: 
`Units.Rotations.of(this / (2 * PI * radius))

src/main/kotlin/frc/robot/lib/ExtensionFunctions.kt
katzuv: I think you can do: 
`this.in(Units.Rotations) * 2 * PI * radius`

Emma03L: I can't I tried

Emma03L: 073b4470ea0f328aaf4d733afae091757ee7d33a

src/main/kotlin/frc/robot/lib/ExtensionFunctions.kt
```diff
@@ -96,3 +100,11 @@ fun Translation2d.flipIfNeeded(): Translation2d =
...
 
 fun Rotation2d.flipIfNeeded(): Rotation2d = if (IS_RED) this.flip() else this
+
+fun Distance.toAngle(radius: Distance): Angle =
+    Units.Rotations.of(
+        this.`in`(Units.Meters) / (2 * PI * radius.`in`(Units.Meters))
+    )
+
+fun Angle.toDistance(radius: Distance): Distance =
+    radius.times(this.`in`(Units.Rotations) * 2 * PI)
```
katzuv, line 110:
```suggestion
    radius * this.`in`(Units.Rotations) * 2 * PI
```

katzuv (changes requested):

src/main/kotlin/frc/robot/lib/ExtensionFunctions.kt
Emma03L: 98a6ff14970919ead738efc2d85d782c8a29e422

katzuv (approved):

PR #73, opened 2025-01-18
Update versions

Emma03L:


katzuv (changes requested):
Awesome!
Note there are other vendordeps worth updating. You can check here: https://github.com/wpilibsuite/vendor-json-repo/blob/main/2025/

src/main/kotlin/frc/robot/subsystems/vision/VisionIOLimelight.java
```diff
@@ -62,7 +63,7 @@ public VisionIOLimelight(String name, Supplier<Rotation2d> rotationSupplier) {
     public void updateInputs(VisionIOInputs inputs) {
         // Update connection status based on whether an update has been seen in the last 250ms
         inputs.connected =
-                (RobotController.getFPGATime() - latencySubscriber.getLastChange()) < 250;
+                ((RobotController.getFPGATime() - latencySubscriber.getLastChange()) / 1000) < 250;
```
vichik123, line 66: If you divide by 1000, wouldn't that be in seconds? And then you would be checking if the latency is 250 seconds.

vichik123 (changes requested):

src/main/kotlin/frc/robot/subsystems/vision/VisionIOLimelight.java
rakrakon: 61fff57b115b0659d37b0cfee24b90472d8ab7b9

rakrakon (approved):

src/main/kotlin/frc/robot/subsystems/vision/VisionIOLimelight.java
Emma03L: So we need to open an issue because that's a change from the new template

katzuv: https://github.com/Mechanical-Advantage/AdvantageKit/pull/137

katzuv (approved):

PR #74, opened 2025-01-18
Add geometry classes factories with default values

katzuv:
- Create geometry factories file
- Add `Translation2d` factories
- Add `Rotation2d` factories
- Add `Pose2d` factories
- Add `Translation3d` factories
- Add `Rotation3d` factories
- Add `Pose3d` factories
- Remove `Pose2d` to `Pose3d` extension function because a WPILib alternative exists

This will be used in #75, among others.

### DIFF
--- a/src/main/kotlin/frc/robot/lib/GeometryFactories.kt
+++ b/src/main/kotlin/frc/robot/lib/GeometryFactories.kt
@@ -3,6 +3,7 @@ package frc.robot.lib
 import edu.wpi.first.math.geometry.Pose2d
 import edu.wpi.first.math.geometry.Rotation2d
 import edu.wpi.first.math.geometry.Translation2d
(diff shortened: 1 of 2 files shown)


src/main/kotlin/frc/robot/lib/GeometryFactories.kt
vichik123, line 1: There's a thing you can do to add these methods to the geometry classes themselves as methods. I can't seem to find how, but there is a way.

vichik123 (changes requested):

src/main/kotlin/frc/robot/lib/GeometryFactories.kt
katzuv: Yes, it's called extension functions, but unfortunately, you can't add constructors.

rakrakon (approved):

PR #76, opened 2025-01-19
Add Qodana workflow from main branch

katzuv:


rakrakon (commented):
I'm not really sure how necessary is this. You should probably ask @Emma03L

katzuv:
It'll catch issues you missed that IntelliJ IDEA shows. I think it's beneficial. After this is merged I'll configure it to ignore files which are not ours.

Emma03L (approved):

katzuv:
Thanks!

PR #80, opened 2025-01-19
Use new geometry factories

katzuv:
- Use correct variable for second stage height
- Rename "carriage" to "second stage"
- Fix comment location
- Use new geometry classes
- Spotless


Emma03L (approved):
🤩

PR #96, opened 2025-01-21
Initialize subsystems

Emma03L:


src/main/kotlin/frc/robot/RobotContainer.kt
```diff
@@ -21,6 +21,13 @@ object RobotContainer {
     private val testController = CommandXboxController(2)
 
     private val swerveDrive = frc.robot.swerveDrive
+    private val vision = frc.robot.vision
+    private val Climber = subsystems["CLimber"]
```
katzuv, line 25:
```suggestion
    private val Climber = subsystems["Climber"]
```

katzuv (changes requested):
We had a more compact way to do that in https://Galaxia5987/Recode2024.

src/main/kotlin/frc/robot/RobotContainer.kt
Emma03L: 37969ef6bdc6cc64a64230d94e51c918a587da5c

src/main/kotlin/frc/robot/RobotContainer.kt
rakrakon, line 1: I would also add the actual Initialization of the subsystems in the RobotContainer.

katzuv:
Try to run IntelliJ IDEA's "Optimize Imports" feature. It might shorten the long imports list.

src/main/kotlin/frc/robot/RobotContainer.kt
Emma03L: It's already there

katzuv (approved):
looks good!

PR #99, opened 2025-01-21
Kotlin translation

Emma03L:


rakrakon (commented):
In my opinion this isn't really a good time for a refactor to the sim classes. I think it's just fine as is for this season especially since everything is already working in the simulation.

katzuv:
> In my opinion this isn't really a good time for a refactor to the sim classes. I think it's just fine as is for this season especially since everything is already working in the simulation.

I think I agree with Barak. BTW, I think the sim classes can be removed, I'm not sure they have an advantage over CTRE's sim classes.

PR #100, opened 2025-01-21
Roller hotfix

Emma03L:


src/main/kotlin/frc/robot/subsystems/intake/roller/RollerConstants.kt
```diff
@@ -3,8 +3,8 @@ package frc.robot.subsystems.intake.roller
 import edu.wpi.first.units.Units
 import edu.wpi.first.units.measure.MomentOfInertia
 
-const val INTAKE_POWER = 0.0
-const val OUTTAKE_POWER = 0.0
-const val FAR_OUTTAKE_POWER = 0.0
+val INTAKE_VOLTAGE = Units.Volts.of(0.0)
+val OUTTAKE_VOLTAGE = Units.Volts.of(0.0)
+val FAR_OUTTAKE_VOLTAGE = Units.Volts.of(0.0)
```
vichik123, line 8: Why is it not a const?

Emma03L: Measures can't be const, it's only for primitive types

PR #109, opened 2025-01-23
Elevator limit switch

Emma03L:


katzuv (changes requested):
I think an input for the limit switch should be included, under the elevator subtable.

Emma03L:
> I think an input for the limit switch should be included, under the elevator subtable.

57f496cc8140f8a1a20d7ea9518e572691df26c0

katzuv (approved):

vichik123 (commented):
Why are there changes in the extender subsystem?

src/main/kotlin/frc/robot/subsystems/intake/extender/ExtenderPorts.kt
```diff
@@ -1,3 +1,3 @@
 package frc.robot.subsystems.intake.extender
 
-const val MOTOR_ID = 0
+const val EXTENDER_MOTOR_ID = 0
```
katzuv, line 3: I think it's better instead to revert the previous name, then in the elevator do `import frc.robot.subsystems.intake.extender.MOTOR_ID as EXTENDER_MOTOR_ID`.

Emma03L:
> Why are there changes in the extender subsystem?

The limit switch is connected to the extender motor because it's the nearest TalonFX

src/main/kotlin/frc/robot/subsystems/intake/extender/ExtenderPorts.kt
Emma03L: 499cf2cc45b188bf0df0497d1481ff4c68255735

PR #113, opened 2025-01-23
Add tunable numbers

rakrakon:
- **Add tunable elevator height**
- **Spotless**
- **Add extender tuning**
- **Add wrist tuning**
- **Add climb angle tuning**
- **Add voltage tuning in intake Roller**
- **Add `.withName()` + Spotless**
- **Fix logging directories**


src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
```diff
@@ -39,6 +43,10 @@ class Elevator(private val io: ElevatorIO) : SubsystemBase() {
     fun l3Algae(): Command = setPosition(Positions.L3_ALGAE)
     fun feeder(): Command = setPosition(Positions.FEEDER)
     fun zero(): Command = setPosition(Positions.ZERO)
+    fun tuningPosition(): Command =
+        runOnce { Positions.TUNING.value = Units.Meters.of(tuningHeight.get()) }
+            .andThen(setPosition(Positions.TUNING))
+            .withName("Elevator/Tuning")
```
katzuv, line 49: I think it's fine to just set the height directly to the IO, there is no reason to pass through the enum.

src/main/kotlin/frc/robot/subsystems/climber/Climber.kt
```diff
@@ -69,6 +73,10 @@ class Climber(private val io: ClimberIO) : SubsystemBase() {
             )
             .withName("climber/setVoltage")
 
+    fun setTuningAngle(): Command =
+        setAngle(Units.Degrees.of(tuningAngleDegrees.get()))
+            .withName("Climb/Tuning")
```
katzuv, line 78: Won't it be more convenient to make all tuning commands continuous? It'll quickly become annoying to need to hit the button every time.

katzuv (changes requested):

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
rakrakon: Yeah, but it's quite a lot to change. Is it that critical if i already did it like this?

src/main/kotlin/frc/robot/subsystems/climber/Climber.kt
rakrakon: So just add something like `.repeatedly()` or `run {command}`?

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
katzuv: Well, if we both agree it's a better solution, then I think it should be amended. I don't think it's a lot to change, you just need to invoke the IO's method and remove the `runOnce()`. It will also return the enum to var, which is better.

Remove you can always consult me with your solution idea before coding, I'd be happy to help!

src/main/kotlin/frc/robot/subsystems/climber/Climber.kt
katzuv: Yes. I guess `run()` is better here, I'd use `repeatedly()` if I already have a `runOnce()` command that I want to run continuously, but this is not the case here.

rakrakon: I'll just use run when i do the comment above and it'll solve both of them.

rakrakon: a3b090d771e19bba4714ed92b288e66a19e170a8

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
rakrakon: a3b090d771e19bba4714ed92b288e66a19e170a8

Emma03L (approved):

PR #118, opened 2025-01-23
Remove compressor

katzuv:


Emma03L:
If I remember correctly there was another place where the compressor was used

Emma03L (approved):

PR #119, opened 2025-01-23
3D-Realistic

DavidMasin:


files/cad/Robot_Final2025Bot/config.json
```diff
@@ -0,0 +1,342 @@
...
+        }
+      ],
+      "zeroedPosition": [
+        -0.0,
+        0,
+        -0.0
+      ]
+    },
+    {
+      "_comment": "model_15 : Gripper  -Roller Algea",
```
katzuv, line 301:
```suggestion
      "_comment": "model_15 : Gripper  -Roller Algae",
```

src/main/kotlin/frc/robot/Robot.kt
```diff
@@ -103,6 +105,7 @@ object Robot : LoggedRobot() {
      */
     override fun robotPeriodic() {
         CommandScheduler.getInstance().run()
+        Logger.recordOutput("ZeroedPosition",Pose3d(0.0,0.0,0.0, Rotation3d(0.0,0.0,0.0)))
```
katzuv, line 108: Should be deleted before merge

src/main/kotlin/frc/robot/subsystems/Visualizer.kt
```diff
@@ -10,88 +9,225 @@ import frc.robot.lib.getRotation3d
...
+        x = Meters.of(-0.25448),
+        y = Meters.of(-0.25448),
+        z = Meters.of(0.0508)
+    )
+private val INITIAL_MODULE_3_Drive_TRANSLATION =
+    getTranslation3d(
+        x = Meters.of(-0.25448),
+        y = Meters.of(-0.25448),
+        z = Meters.of(0.0508)
+    )
```
katzuv, line 62: You can make a list based on the distances of the modules from the center of rotation.
Also, if you need to specify all three axes, just do `Translation3d(1,2,3)`.

src/main/kotlin/frc/robot/subsystems/Visualizer.kt
```diff
@@ -10,88 +9,225 @@ import frc.robot.lib.getRotation3d
...
-                    getTranslation3d(z = secondStageHeight)
+                x = INITIAL_Elevator_2_TRANSLATION.x,
+                z = secondStageHeight.plus(INITIAL_Elevator_2_TRANSLATION.z)
             )
         return Pair(firstStagePose, secondStagePose)
     }
 
     @AutoLogOutput
     fun getSubsystemsPoses(): Array<Pose3d> {
+        val Module0PoseTurn =
```
katzuv, line 111: Use a function to get the pose for each module, then another function that'll return the poses of all modules.

src/main/kotlin/frc/robot/subsystems/Visualizer.kt
```diff
@@ -10,88 +9,225 @@ import frc.robot.lib.getRotation3d
...
+    getTranslation3d(x = Meters.of(-0.24), z = Meters.of(0.285))
+
+class Visualizer(){
+    private val swerveDrive = frc.robot.swerveDrive
+    private val climber = frc.robot.climber
+    private val elevator = frc.robot.elevator
+    private val gripper = frc.robot.gripper
+    private val extender = frc.robot.extender
+    private val roller = frc.robot.roller
+    private val wrist = frc.robot.wrist
```
katzuv, line 91: Why is this change needed?

src/main/kotlin/frc/robot/subsystems/Visualizer.kt
```diff
@@ -10,88 +9,225 @@ import frc.robot.lib.getRotation3d
...
+                getTranslation3d(
+                    x = INITIAL_WRIST_TRANSLATION.x,
+                    y = INITIAL_WRIST_TRANSLATION.y,
+                    z =
+                        INITIAL_WRIST_TRANSLATION.z.plus(
+                            elevator.height.invoke().`in`(Meters)
+                        )
+                ),
+                getRotation3d(pitch = -wrist.angle.invoke())
             )
```
katzuv, line 194: Use regular constructors of the `Pose3d` and `Translation3d` classes.

src/main/kotlin/frc/robot/subsystems/Visualizer.kt
```diff
@@ -10,88 +9,225 @@ import frc.robot.lib.getRotation3d
...
-
-        val coralRollersPose =
-            wristPose.rotateBy(
-                getRotation3d(pitch = coralRollersAngle.invoke())
+            getPose3d(
+                getTranslation3d(
+                    x = INITIAL_WRIST_TRANSLATION.x,
+                    y = INITIAL_WRIST_TRANSLATION.y,
+                    z =
+                        INITIAL_WRIST_TRANSLATION.z.plus(
```
katzuv, line 189: You can do `+`

src/main/kotlin/frc/robot/subsystems/drive/Drive.java
```diff
@@ -205,6 +207,17 @@ public void periodic() {
...
         odometryLock.unlock();
+        SwerveTurnAngle[0] = modules[0].getAngle().getMeasure();
+        SwerveTurnAngle[1] = modules[1].getAngle().getMeasure();
+        SwerveTurnAngle[2] = modules[2].getAngle().getMeasure();
+        SwerveTurnAngle[3] = modules[3].getAngle().getMeasure();
+
+        SwerveDriveAngle[0] = Radians.of(modules[0].getWheelRadiusCharacterizationPosition());
+        SwerveDriveAngle[1] = Radians.of(modules[1].getWheelRadiusCharacterizationPosition());
+        SwerveDriveAngle[2] = Radians.of(modules[2].getWheelRadiusCharacterizationPosition());
+        SwerveDriveAngle[3] = Radians.of(modules[3].getWheelRadiusCharacterizationPosition());
```
katzuv, line 218: I don't think you want to use the characterization function

src/main/kotlin/frc/robot/subsystems/vision/Vision.java
katzuv, line 1: Should be reverted

katzuv (changes requested):

src/main/kotlin/frc/robot/subsystems/Visualizer.kt
DavidMasin: Why? 
What's wrong with getPose3d? and getTranslation3d? and getRotation3d?

src/main/kotlin/frc/robot/subsystems/drive/Drive.java
DavidMasin: Why?
I want the wheel position in rad. According to the documentation it says: 
Returns the module position in radians
Is it wrong?

src/main/kotlin/frc/robot/subsystems/Visualizer.kt
DavidMasin: Previously it all was in the Robot container and the Visualizer class got a lot of variables. I just moved it here

I just need to know information about the subsystems.

src/main/kotlin/frc/robot/subsystems/Visualizer.kt
rakrakon: b9a11da517e62d47b4c4f8409889a6a1c537f509

src/main/kotlin/frc/robot/subsystems/drive/Drive.java
katzuv: OK nevermind. It's just the name of the function is weird.

src/main/kotlin/frc/robot/subsystems/Visualizer.kt
katzuv: There are still usages of `plus()`

src/main/kotlin/frc/robot/subsystems/Visualizer.kt
katzuv: I added these [factory functions](https://en.wikipedia.org/wiki/Factory_(object-oriented_programming)) because the Java classes don't support default parameters, so even if you only want to pass one value, you need to pass all three.
But if you anyway have to pass all three, using the regular constructors is better.
Note that in this case I believe you can do:
```kotlin
        val wristPose =
            getPose3d(
                INITIAL_WRIST_TRANSLATION + getTranslation3d(z = elevator.height.invoke()),
                getRotation3d(pitch = wrist.angle.invoke())
            )
```
But I didn't test this because my laptop is a 🥔.

src/main/kotlin/frc/robot/subsystems/Visualizer.kt
katzuv: In most subsystems you only need to a single value, there is no need to pass the whole subsystem.

src/main/kotlin/frc/robot/subsystems/Visualizer.kt
```diff
@@ -1,97 +1,156 @@
...
             getPose3d(
-                INITIAL_ELEVATOR_TRANSLATION +
-                    getTranslation3d(z = secondStageHeight)
+                x = INITIAL_Elevator_2_TRANSLATION.x,
+                z = secondStageHeight.plus(INITIAL_Elevator_2_TRANSLATION.z)
             )
         return Pair(firstStagePose, secondStagePose)
     }
 
+    private fun getSwerveModulePose(): Pair<Array<Pose3d>, Array<Pose3d>> {
```
katzuv, line 61: It's better to process a single module, and then have the for loop somewhere else (in another function if you prefer).

src/main/kotlin/frc/robot/subsystems/Visualizer.kt
```diff
@@ -1,97 +1,156 @@
...
+                        0.0508
+                    ),
+                    getRotation3d(yaw = swerveDrive.SwerveTurnAngle[i])
+                )
+            swervePosesDrive[i] =
+                Pose3d(
+                    Translation3d(
+                        swerveMoudlePose[i].x,
+                        swerveMoudlePose[i].y,
+                        0.0508
```
katzuv, line 81: Both should be constants.

src/main/kotlin/frc/robot/subsystems/Visualizer.kt
katzuv, line 61: It's better to process a single module, and then have the for loop somewhere else (in another function if you prefer).

katzuv (changes requested):
Use the functions added in #144 after it's merged.

src/main/kotlin/frc/robot/subsystems/Visualizer.kt
DavidMasin: Are you talking about the 0.0508?

katzuv: Yes

DavidMasin: https://github.com/Galaxia5987/robot-2025/pull/119/commits/bbf4dd4c8896ae89a5ba42d199d41bd84ee88a9a

src/main/kotlin/frc/robot/subsystems/Visualizer.kt
DavidMasin: https://github.com/Galaxia5987/robot-2025/pull/119/commits/bdede27a0ee9e60a722ec9b52ac4ba389e5aaafe

src/main/kotlin/frc/robot/subsystems/Visualizer.kt
katzuv: The best solution is to revert to how it's been, where you inject the visualizer the required data. I want us (after we merge this PR) to be able to visualize the subsystems setpoints to, so we need to be able to pass different data.

DavidMasin: https://github.com/Galaxia5987/robot-2025/pull/119/commits/6057a78514fb77b0b69789beb9d0d64346cc8cfd
Like this?

katzuv: Not what I originally meant but it's good

katzuv (approved):
Awesome!

Emma03L (approved):

PR #122, opened 2025-01-24
Intake safety feature

Emma03L:


src/main/kotlin/frc/robot/subsystems/intake/extender/Extender.kt
```diff
@@ -29,14 +31,22 @@ class Extender(private val io: ExtenderIO) : SubsystemBase() {
...
-        runOnce {
-                io.setPosition(position.position)
-                setpoint = position.position
-                setpointName = position.getLoggingName()
-            }
+    private fun setPosition(position: () -> Distance): Command =
+        sequence(
+                runOnce {
+                    setpoint = position.invoke()
+                    io.setPosition(position.invoke())
```
katzuv, line 38:
```suggestion
                    setpoint = position.invoke()
                    io.setPosition(setpoint)
```
While this probably won't be a problem in this case. it's better not to invoke a supplier twice, because if the value changes between the two invocations, you might get different values.

src/main/kotlin/frc/robot/subsystems/intake/extender/Extender.kt
```diff
@@ -92,7 +102,8 @@ class Extender(private val io: ExtenderIO) : SubsystemBase() {
     @AutoLogOutput
     private var atSetpoint = Trigger {
         io.inputs.position.isNear(setpoint, POSITION_TOLERANCE)
-    }
+    }.negate().debounce(SAFETY_DEBOUNCE)
+        .onTrue(setPosition { setpoint })
```
katzuv, line 106: I'm not sure this will have the intended effect. Just because the command ends, it doesn't mean the motor's control request will now be different.

katzuv (changes requested):

src/main/kotlin/frc/robot/subsystems/intake/extender/Extender.kt
katzuv: Never mind I missed the set=zero.

katzuv (approved):

src/main/kotlin/frc/robot/subsystems/intake/extender/Extender.kt
Emma03L: The setpoint variable is always the most recent setpoint, the only way it changes is when a different command is called at which point it the new setpoint is the one we want the extender to return to if it gets hit.

Emma03L: I now realized that adding the logic at the end of the trigger changes the trigger so it has to be a default command

PR #124, opened 2025-01-24
Add soft limits

Emma03L:


src/main/kotlin/frc/robot/subsystems/intake/extender/Extender.kt
```diff
@@ -55,7 +55,10 @@ class Extender(private val io: ExtenderIO) : SubsystemBase() {
 
     fun reset(): Command {
         return setVoltage(RESET_VOLTAGE)
-            .alongWith(runOnce { finishedResettingFlag = false })
+            .alongWith(
+                runOnce { finishedResettingFlag = false },
+                runOnce { io.setSoftLimits(false) }
```
katzuv, line 60: I'd put both inside the same `runOnce()`

katzuv (changes requested):

PR #129, opened 2025-01-26
Driver GUI NetworkTables

rakrakon:
- **Create Pose Constants**
- **Create `getBranchCommand`**
- **Create ReefDriveCommands & Spotless**


src/main/kotlin/frc/robot/ReefDriveCommands.kt
```diff
@@ -0,0 +1,25 @@
...
+        7 to Commands.none(), // TODO: Fill in actual command
+        8 to Commands.none(), // TODO: Fill in actual command
+        9 to Commands.none(), // TODO: Fill in actual command
+        10 to Commands.none(), // TODO: Fill in actual command
+        11 to Commands.none(), // TODO: Fill in actual command
+        12 to Commands.none(), // TODO: Fill in actual command (Coral Station)
+        13 to Commands.none() // TODO: Fill in actual command (Coral Station)
+    )
+
+fun getReefDriveCommand(): Command = REEF_MAP[NetworkTablesJNI.getEntry(NetworkTablesJNI.getDefaultInstance(), "Dashboard/TargetReefPose")]!!
```
katzuv, line 25: Don't use JNI stuff. See how AdvantageKit does that:
https://github.com/Mechanical-Advantage/AdvantageKit/blob/main/akit/src/main/java/org/littletonrobotics/junction/networktables/LoggedNetworkNumber.java

src/main/kotlin/frc/robot/PoseConstants.kt
```diff
@@ -0,0 +1,19 @@
...
+val REEF_SIX_POSE = Pose3d() // TODO: Fill in
+val REEF_SEVEN_POSE = Pose3d() // TODO: Fill in
+val REEF_EIGHT_POSE = Pose3d() // TODO: Fill in
+val REEF_NINE_POSE = Pose3d() // TODO: Fill in
+val REEF_TEN_POSE = Pose3d() // TODO: Fill in
+val REEF_ELEVEN_POSE = Pose3d() // TODO: Fill in
+val REEF_TWELVE_POSE = Pose3d() // TODO: Fill in
+
+val LEFT_FEEDER_POSE = Pose3d() // TODO: Fill in
+val RIGHT_FEEDER_POSE = Pose3d() // TODO: Fill in
```
katzuv, line 19: 1. Why not make this a map?
2. I think it's better to load these from Choreo's variables, it's better to have a single source of truth. You can also calculate the locations from the Reef's center.

src/main/kotlin/frc/robot/ReefDriveCommands.kt
```diff
@@ -0,0 +1,25 @@
...
+        5 to Commands.none(), // TODO: Fill in actual command
+        6 to Commands.none(), // TODO: Fill in actual command
+        7 to Commands.none(), // TODO: Fill in actual command
+        8 to Commands.none(), // TODO: Fill in actual command
+        9 to Commands.none(), // TODO: Fill in actual command
+        10 to Commands.none(), // TODO: Fill in actual command
+        11 to Commands.none(), // TODO: Fill in actual command
+        12 to Commands.none(), // TODO: Fill in actual command (Coral Station)
+        13 to Commands.none() // TODO: Fill in actual command (Coral Station)
+    )
```
katzuv, line 23: Do we really need to have a command for each branch? I think it's fine to have a single command that by default pulls the target branch from NetworkTables, but can also receive a hard coded value. Unless, you intend to use these commands in autonomous, too.

src/main/kotlin/frc/robot/RobotContainer.kt
```diff
@@ -33,6 +36,15 @@ object RobotContainer {
     private val wrist = frc.robot.wrist
     val visualizer: Visualizer
 
+    private val BRANCH_MAP =
```
katzuv, line 39: Height map?

src/main/kotlin/frc/robot/RobotContainer.kt
```diff
@@ -33,6 +36,15 @@ object RobotContainer {
...
+        mapOf(
+            1 to l1(Trigger { true }), // TODO: Fill in correct trigger
+            2 to l2(Trigger { true }), // TODO: Fill in correct trigger
+            3 to l3(Trigger { true }), // TODO: Fill in correct trigger
+            4 to l4(Trigger { true }), // TODO: Fill in correct trigger
+            5 to feeder(Trigger { true }) // TODO: Fill in correct trigger
+        )
```
katzuv, line 46: I think it's better to put that in the previous file and rename it

katzuv (changes requested):

src/main/kotlin/frc/robot/PoseConstants.kt
```diff
@@ -0,0 +1,19 @@
...
+val REEF_THREE_POSE = Pose3d() // TODO: Fill in
+val REEF_FOUR_POSE = Pose3d() // TODO: Fill in
+val REEF_FIVE_POSE = Pose3d() // TODO: Fill in
+val REEF_SIX_POSE = Pose3d() // TODO: Fill in
+val REEF_SEVEN_POSE = Pose3d() // TODO: Fill in
+val REEF_EIGHT_POSE = Pose3d() // TODO: Fill in
+val REEF_NINE_POSE = Pose3d() // TODO: Fill in
+val REEF_TEN_POSE = Pose3d() // TODO: Fill in
+val REEF_ELEVEN_POSE = Pose3d() // TODO: Fill in
+val REEF_TWELVE_POSE = Pose3d() // TODO: Fill in
```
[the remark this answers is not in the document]
Emma03L: I don't think so, you should use the convention we use for auto routines. 1 thorough 6 with left and right to indicate which pole.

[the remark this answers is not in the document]
katzuv: I think we should change ours. Otherwise, there might be confusion with other teames in the competition: https://www.chiefdelphi.com/t/naming-convention-for-the-reef/479883/11?u=dan

src/main/kotlin/frc/robot/ReefDriveCommands.kt
rakrakon: Why not? It's shorter.
Also the NetworkTableInstance class is just a wrapper for the NetworkTables JNI class

katzuv: Right, use the wrapper... Don't use stuff you're not supposed to use. It's a Good Practice.

src/main/kotlin/frc/robot/PoseConstants.kt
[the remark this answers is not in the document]
rakrakon: 16f0cc8
This is the convention we agreed on.

src/main/kotlin/frc/robot/ReefDriveCommands.kt
rakrakon: b8b34e0

src/main/kotlin/frc/robot/PoseConstants.kt
rakrakon: This isn't a map because this will be passed as a parameter to the `moveAutonomously` command.

src/main/kotlin/frc/robot/ReefDriveCommands.kt
rakrakon: Sure, i'll just make this return the pose

src/main/kotlin/frc/robot/PoseConstants.kt
katzuv: Why can't you do `moveAutonomously(POSES[2L])` or something similar? 
I'd also find a shorter name for this command.

src/main/kotlin/frc/robot/PoseConstants.kt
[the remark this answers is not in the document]
katzuv: Good luck I'm strategy meetings…

src/main/kotlin/frc/robot/PoseConstants.kt
rakrakon: That happens in the `ReefDriveCommands`. Currently it's not implemented because we haven't setup the autoDrive command.
8a10085
aeffd57

src/main/kotlin/frc/robot/ReefDriveCommands.kt
rakrakon: 8a10085

src/main/kotlin/frc/robot/RobotContainer.kt
rakrakon: 5945e09

src/main/kotlin/frc/robot/RobotContainer.kt
rakrakon: We can't put that in the previous file because the corresponding command needs to take the end Triggers as a parameter.

katzuv:
Perhaps we should have two sets of poses for both alliances. They might not be entirely symmetrical in the real field.

src/main/kotlin/frc/robot/PoseConstants.kt
katzuv, line 1: Processor is missing?

rakrakon: There is no Processor in the GUI.

rakrakon:
> Perhaps we should have two sets of poses for both alliances. They might not be entirely symmetrical in the real field.

There is flipping utils so we won't need to do that.

katzuv:
> > Perhaps we should have two sets of poses for both alliances. They might not be entirely symmetrical in the real field.
> 
> There is flipping utils so we won't need to do that.

Read again what I wrote.

rakrakon:
> > > Perhaps we should have two sets of poses for both alliances. They might not be entirely symmetrical in the real field.
> > 
> > 
> > There is flipping utils so we won't need to do that.
> 
> Read again what I wrote.

Why would they not be symetrical?

katzuv:
Bad tolerances while assembling the field + robots moving the field elements during match play.

rakrakon:
> Bad tolerances while assembling the field + robots moving the field elements during match play.

I don't see how we can know how much does the field element moves by having a set of poses for each team.

katzuv:
Measure it :)
That'll help at least in the first morning.

rakrakon:
> Measure it :) That'll help at least in the first morning.

You mean manually?

src/main/kotlin/frc/robot/PoseConstants.kt
katzuv: Ok so let's just not forget to add it later.

src/main/kotlin/frc/robot/RobotContainer.kt
katzuv: If all the triggers are the same we can have a single command in the other file that'll do the map thing?

rakrakon: Yes, but they won't be the same since we might want to use this in auto

Emma03L (approved):

src/main/kotlin/frc/robot/RobotContainer.kt
```diff
@@ -33,6 +34,15 @@ object RobotContainer {
     private val wrist = frc.robot.wrist
     val visualizer: Visualizer
 
+    private val HEIGHT_MAP =
```
katzuv, line 37: This doesn't map the height, it has commands as values.

src/main/kotlin/frc/robot/lib/network_tables/NetworkTable.kt
```diff
@@ -0,0 +1,12 @@
+package frc.robot.lib.network_tables
+
+class NetworkTable(private val io: NetworkTableIO) {
+
+    fun getDoubleFromTopic(topicName: String): Double =
+        io.getDoubleFromTopic(topicName)
```
katzuv, line 6: This is not replayable.

katzuv (changes requested):

src/main/kotlin/frc/robot/ReefDriveCommands.kt
```diff
@@ -0,0 +1,30 @@
...
+        6 to REEF_6L_POSE, // TODO: Fill in actual command
+        7 to REEF_1R_POSE, // TODO: Fill in actual command
+        8 to REEF_1L_POSE, // TODO: Fill in actual command
+        9 to REEF_2R_POSE, // TODO: Fill in actual command
+        10 to REEF_2L_POSE, // TODO: Fill in actual command
+        11 to REEF_3R_POSE, // TODO: Fill in actual command
+        12 to RIGHT_FEEDER_INNER_POSE, // TODO: Fill in actual command (Coral Station)
+        13 to RIGHT_FEEDER_OUTER_POSE, // TODO: Fill in actual command (Coral Station)
+        14 to LEFT_FEEDER_INNER_POSE, // TODO: Fill in actual command (Coral Station)
+        15 to LEFT_FEEDER_OUTER_POSE // TODO: Fill in actual command (Coral Station)
```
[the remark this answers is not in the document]
Emma03L: But then it would be `LEFT_FEEDER_LEFT_POSE` which sounds weird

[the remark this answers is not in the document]
katzuv: Maybe weird, but clearer. Maybe do left/right-close/far, from the alliance station POV.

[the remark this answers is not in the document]
Emma03L: fe7ced1ea08e7f65a6334a4beea46cd060f2f486

src/main/kotlin/frc/robot/ReefDriveCommands.kt
```diff
@@ -0,0 +1,30 @@
...
+    )
+
+private fun getDriveToPose(): Pose3d =
+    REEF_MAP[networkTables.getIntFromTopic("/Dashboard/TargetReefPose")]!!
+
+fun getDriveToCommand(): Command = Commands.none() // TODO: Implement
```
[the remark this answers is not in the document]
Emma03L: 07d848c1fef80d4e63590ec789362a70461b6bdd

src/main/kotlin/frc/robot/ReefDriveCommands.kt
```diff
@@ -0,0 +1,30 @@
...
+        10 to REEF_2L_POSE, // TODO: Fill in actual command
+        11 to REEF_3R_POSE, // TODO: Fill in actual command
+        12 to RIGHT_FEEDER_INNER_POSE, // TODO: Fill in actual command (Coral Station)
+        13 to RIGHT_FEEDER_OUTER_POSE, // TODO: Fill in actual command (Coral Station)
+        14 to LEFT_FEEDER_INNER_POSE, // TODO: Fill in actual command (Coral Station)
+        15 to LEFT_FEEDER_OUTER_POSE // TODO: Fill in actual command (Coral Station)
+    )
+
+private fun getDriveToPose(): Pose3d =
+    REEF_MAP[networkTables.getIntFromTopic("/Dashboard/TargetReefPose")]!!
```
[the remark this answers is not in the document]
Emma03L: f794c47c41fefcf198e5be9bea8238a34cd9760d

src/main/kotlin/frc/robot/Initializer.kt
```diff
@@ -152,3 +156,12 @@ val wrist =
                 }
         }
     )
+
+val networkTables =
+    NetworkTable(
+        when (CURRENT_MODE) {
+            Mode.REAL -> NetworkTableIOReal()
+            Mode.SIM -> NetworkTableIOSim()
```
katzuv, line 164: Did you test this code with the GUI used, while in simulation?

src/main/kotlin/frc/robot/RobotContainer.kt
Emma03L: e61769f177d499e5c4d6e960d7177e3cdf30b360

src/main/kotlin/frc/robot/lib/network_tables/NetworkTable.kt
rakrakon: I'm a bit confused on why this needs to be replayable. This controls only the commands so since there's no inputs and only commands this should be fine if I'm not mistaken.

katzuv: Think what this function will return during replay. You can also run replay and see what happens.

src/main/kotlin/frc/robot/Initializer.kt
rakrakon: Yes

src/main/kotlin/frc/robot/RobotContainer.kt
```diff
@@ -76,13 +79,48 @@ object RobotContainer {
...
+            .a()
+            .onTrue(
+                Commands.defer(
+                    {
+                        createScoreCommandMap()[
+                            networkTables.getTargetBranchPose()]
+                    },
+                    setOf(gripper, wrist, elevator)
+                )
+            )
```
Emma03L, line 122: Why is this here?

src/main/kotlin/frc/robot/auto/network_tables/NetworkTableIO.kt
```diff
@@ -0,0 +1,15 @@
...
+
+interface NetworkTableIO {
+    var inputs: LoggedNetworkTableInputs
+
+    fun updateInputs() {}
+
+    @Logged
+    open class NetworkTableInputs {
+        var targetBranchPose: Int = 0
+        var targetReefPose: Int = 0
```
katzuv, line 13: If this includes the coral station location the name should change. Why not make two topics though? Now if the driver selected a coral station last the robot code will have no way the know the desired reef pose

katzuv (changes requested):

PR #130, opened 2025-01-26
Add SysID

Emma03L:


src/main/kotlin/frc/robot/lib/ExtensionFunctions.kt
```diff
@@ -111,6 +113,24 @@ fun Angle.toDistance(radius: Distance, gearRatio: Double): Distance =
...
+            .of(1.0 / (radius.`in`(Units.Meters) * gearRatio * 2.0 * PI))
+    )
+
+fun AngularVelocity.toLinear(
+    radius: Distance,
+    gearRatio: Double
+): LinearVelocity =
+    this.timesConversionFactor(
+        Units.MetersPerSecond.per(Units.RotationsPerSecond)
+            .of(radius.`in`(Units.Meters) * gearRatio * 2.0 * PI)
```
katzuv, line 131: I think it's better to put the `per`s in constants, otherwise they'll be created again each time. Should be changed in the previous extensions as well.

src/main/kotlin/frc/robot/subsystems/climber/ClimberIO.kt
```diff
@@ -14,6 +16,7 @@ interface ClimberIO {
     fun setAngle(angle: Angle) {}
     fun setStopperPower(power: Double) {}
     fun updateInput() {}
+    fun updateRoutineLog(log: SysIdRoutineLog) {}
```
katzuv, line 19: I don't think this is needed. You should use CTRE's signal logger: https://v6.docs.ctr-electronics.com/en/2024/docs/api-reference/wpilib-integration/sysid-integration/index.html#advantages-of-signallogger-over-datalog

katzuv (changes requested):

src/main/kotlin/frc/robot/lib/ExtensionFunctions.kt
Emma03L: They use parameters so They have to be inside the function

src/main/kotlin/frc/robot/subsystems/climber/ClimberIO.kt
rakrakon: If i'm not mistaken as long as we log `position`, `velocity` & `voltage` there's no need for this function at all.

Emma03L: I logged the state with the advantage kit logger 08e7865152ba9f662ff4c33d1022738c822105f3

katzuv (approved):

PR #131, opened 2025-01-26
Disable commands

Emma03L:


src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
```diff
@@ -74,6 +74,8 @@ class Elevator(private val io: ElevatorIO) : SubsystemBase() {
             .andThen(runOnce(io::reset))
             .withName("Elevator/reset")
 
+    fun disableCommand(): Command = setVoltage(Units.Volts.zero())
```
rakrakon, line 77: The `Units.Volts.zero()` should be extracted into a Constant

src/main/kotlin/frc/robot/subsystems/intake/extender/Extender.kt
```diff
@@ -88,6 +89,11 @@ class Extender(private val io: ExtenderIO) : SubsystemBase() {
             .onTrue(setPosition { setpoint })
     }
 
+    fun disableCommand(): Command =
```
rakrakon, line 92: I think it's better to do this as a trigger. Then you won't need the `until()`

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
katzuv, line 77: Isn't it better to hold the current height?

src/main/kotlin/frc/robot/lib/ExtensionFunctions.kt
```diff
@@ -97,3 +100,7 @@ fun CommandXboxController.setRumble(strength: Double) {
 fun CommandXboxController.rumbleCommand(): Command {
     return Commands.startEnd({ this.setRumble(1.0) }, { this.setRumble(0.0) })
 }
+
+fun Subsystem.setDisableCommand(command: Command) {
+    val isDisabled = Trigger { DriverStation.isDisabled() }.whileTrue(command)
```
katzuv, line 105: Nice!
You should add an optional delay time for the drivetrain (which is missing a disable command). Note also WPILib has a built-in trigger for disable you should use.

rakrakon (changes requested):

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
katzuv: I'd say `zero()` is already a constant.

katzuv (changes requested):
Wrist is missing?

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
Emma03L: Yeah I agree with Katzuv

Emma03L: Not necessarily. That's why I said it really depends on the situation. Sometimes we'll want to disable it and move it, other times we'll want it to stay in place...

src/main/kotlin/frc/robot/lib/ExtensionFunctions.kt
Emma03L: I'll debounce the trigger

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
katzuv: I think it should always hold position. Otherwise it might fall down and that'll be a problem if the wrist is open.

katzuv: Perhaps we need to understand which we'll need _more_.

src/main/kotlin/frc/robot/lib/ExtensionFunctions.kt
Emma03L: I didn't find the WPILib disable trigger

katzuv: https://github.wpilib.org/allwpilib/docs/development/java/edu/wpi/first/wpilibj2/command/button/RobotModeTriggers.html
You can also find that in https://github.com/Galaxia5987/Recode2024/pull/43 (which should also be implemented).

PR #153, opened 2025-02-03
Calibrate Intake

rakrakon:
- **Add constants for canbus names**
- **Oops (used same port for two motors)**
- **Update offsets**
- **Calibrate PID**
- **Calibrate slip current**
- **Calibrate max velocity**
- **Use mxp for navx**
- **Calibrate omega PID**
- **Apply spotless**
- **Update ports**
- **Calibrate constants**
- **Change neutral mode to break**
- **Calibrate PID**
- **Fix tuning position**
- **Calibrate softlimits & pid**
- **Switch to use `PositionVoltage`**
- **Increase INTAKE_VOLTAGE**

### DIFF
--- a/src/main/kotlin/frc/robot/subsystems/intake/extender/ExtenderConstants.kt
+++ b/src/main/kotlin/frc/robot/subsystems/intake/extender/ExtenderConstants.kt
@@ -1,10 +1,7 @@
 package frc.robot.subsystems.intake.extender
 
 import edu.wpi.first.units.Units
-import edu.wpi.first.units.measure.Current
-import edu.wpi.first.units.measure.Distance
-import edu.wpi.first.units.measure.MomentOfInertia
-import edu.wpi.first.units.measure.Voltage
+import edu.wpi.first.units.measure.*
 import frc.robot.lib.Gains
 import frc.robot.lib.selectGainsBasedOnMode
 
@@ -13,15 +10,15 @@ val RESET_VOLTAGE: Voltage = Units.Volts.of(0.0)
 val PINION_RADIUS: Distance = Units.Millimeters.of(15.22)
 val MOMENT_OF_INERTIA: MomentOfInertia = Units.KilogramSquareMeters.of(0.003)
 val RESET_CURRENT_THRESHOLD: Current = Units.Amps.of(0.0)
-val POSITION_TOLERANCE: Distance = Units.Centimeters.of(0.7)
-val MAX_EXTENSION: Distance = Units.Meters.of(0.0)
-val MIN_EXTENSION: Distance = Units.Meters.of(0.0)
+val POSITION_TOLERANCE: Distance = Units.Centimeters.of(0.1)
+val MAX_EXTENSION: Angle = Units.Rotations.of(6.7)
+val MIN_EXTENSION: Angle = Units.Rotations.of(-6.3)
 const val SAFETY_DEBOUNCE = 1.0
 
-val GAINS = selectGainsBasedOnMode(Gains(1.5), Gains(kP = 0.5, kD = 0.2))
+val GAINS = selectGainsBasedOnMode(Gains(2.0), Gains(kP = 0.5, kD = 0.2))
 
 enum class Positions(val position: Distance) {
-    EXTENDED(Units.Meters.of(0.3)),
+    EXTENDED(Units.Meters.of(0.4)),
     RETRACTED(Units.Meters.of(0.0));
 
     fun getLoggingName() =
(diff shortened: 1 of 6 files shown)


src/main/kotlin/frc/robot/subsystems/intake/extender/ExtenderConstants.kt
```diff
@@ -13,15 +10,15 @@ val RESET_VOLTAGE: Voltage = Units.Volts.of(0.0)
 val PINION_RADIUS: Distance = Units.Millimeters.of(15.22)
 val MOMENT_OF_INERTIA: MomentOfInertia = Units.KilogramSquareMeters.of(0.003)
 val RESET_CURRENT_THRESHOLD: Current = Units.Amps.of(0.0)
-val POSITION_TOLERANCE: Distance = Units.Centimeters.of(0.7)
-val MAX_EXTENSION: Distance = Units.Meters.of(0.0)
-val MIN_EXTENSION: Distance = Units.Meters.of(0.0)
+val POSITION_TOLERANCE: Distance = Units.Centimeters.of(0.1)
+val MAX_EXTENSION: Angle = Units.Rotations.of(6.7)
+val MIN_EXTENSION: Angle = Units.Rotations.of(-6.3)
```
katzuv, line 15: Why?

src/main/kotlin/frc/robot/subsystems/intake/extender/ExtenderIOReal.kt
```diff
@@ -79,6 +75,7 @@ class ExtenderIOReal : ExtenderIO {
                 position.toAngle(PINION_RADIUS, GEAR_RATIO)
             )
         )
+        println(position.toAngle(PINION_RADIUS, GEAR_RATIO))
```
katzuv, line 78: Remove

katzuv (changes requested):

src/main/kotlin/frc/robot/subsystems/intake/extender/ExtenderConstants.kt
rakrakon: Soft limits takes rotations and we calibrated this with Phoenix tuner x.

katzuv: Sure, but you can still use centimeters here and convert the measured to rotations.

src/main/kotlin/frc/robot/subsystems/intake/extender/ExtenderIOReal.kt
rakrakon: 6c7503ec047becd13c4d5d5299a4d17bfa078935

rakrakon: 6c7503ec047becd13c4d5d5299a4d17bfa078935

src/main/kotlin/frc/robot/subsystems/intake/extender/ExtenderConstants.kt
rakrakon: If i use centimeters here i'll have to measure again. Since we measured in rotations via Tuner X i don't see why that is necessary

katzuv: > If i use centimeters here i'll have to measure again. 

Maybe. But in this case you know the gear ratio and everything and you can calculate it. Or, you can take a tape and measure it. Or, you can check the position input in AdvantageScope which is in meters (which is another reason the min/max positions should be in meters). and take the value from there.
All of these solutions will take you less than a minute to implement, probably less time than it takes to discuss the matter.

Emma03L: Yes, we can calculate it but there is no reason. We calibrated it through tuner x so the value will be in rotations. There is no need to convert it to a distance in the constants file and then right back to rotations

katzuv: The reason to do so is to make it easier to change in the future, and to retain the same dimensions -- again, if position is measured as a distance (as it should), it doesn't make sense to write its min/max values in another dimension.

Emma03L: We are measuring the MAX & MIN Position via Phoenix x which measure in rotations.

katzuv: You both said that multiple times and I don't see how this is relevant.

katzuv (approved):

src/main/kotlin/frc/robot/subsystems/intake/extender/ExtenderConstants.kt
rakrakon: So by your logic if for example we can only measure in Inches for some reason. And our code takes Inches, We should measure in inches convert to Centimeters and then convert back to Inches in the code. How is that logical?

PR #155, opened 2025-02-05
Calibrate Elevator

Emma03L:
- **Calibrate PID**
- **Switch to PositionVoltage + Correct CANBUS**
- **Fix `setTuningPosition`**
- **Correct Ports**
- **Fix input updates**
- **Calibrate Constants**
- **Change setHeight to `run`**
- **Fix reset Command**
- **Spotless**

### DIFF
--- a/src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
+++ b/src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
@@ -69,8 +69,7 @@ class Elevator(private val io: ElevatorIO) : SubsystemBase() {
                 setpointValue = height
                 io.setHeight(height)
             }
-                .withName("Elevator/Tuning")
-        }
+        }.withName("Elevator/Tuning")
 
     fun setVoltage(voltage: Voltage): Command =
         startEnd(
--- a/src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
+++ b/src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
@@ -8,8 +8,8 @@ import edu.wpi.first.wpilibj2.command.Command
 import edu.wpi.first.wpilibj2.command.Commands
 import edu.wpi.first.wpilibj2.command.SubsystemBase
 import edu.wpi.first.wpilibj2.command.button.Trigger
-import java.util.function.DoubleSupplier
 import edu.wpi.first.wpilibj2.command.sysid.SysIdRoutine
+import java.util.function.DoubleSupplier
 import org.littletonrobotics.junction.AutoLogOutput
 import org.littletonrobotics.junction.Logger
 import org.littletonrobotics.junction.mechanism.LoggedMechanism2d
@@ -66,12 +66,13 @@ class Elevator(private val io: ElevatorIO) : SubsystemBase() {
 
     fun tuningPosition(): Command =
         defer {
-            run {
-                val height = Units.Meters.of(tuningHeight.get())
-                setpointValue = height
-                io.setHeight(height)
+                run {
+                    val height = Units.Meters.of(tuningHeight.get())
+                    setpointValue = height
+                    io.setHeight(height)
+                }
             }
-        }.withName("Elevator/Tuning")
+            .withName("Elevator/Tuning")
 
     fun setVoltage(voltage: Voltage): Command =
         startEnd(
--- a/src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
+++ b/src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
@@ -63,17 +63,14 @@ class Elevator(private val io: ElevatorIO) : SubsystemBase() {
         setHeight(Positions.ZERO).withName("Elevator/Move To Zero")
 
     fun tuningPosition(): Command =
-        Commands.defer(
-            {
-                run {
-                        val height = Units.Meters.of(tuningHeight.get())
-                        setpointValue = height
-                        io.setHeight(height)
-                    }
-                    .withName("Elevator/Tuning")
-            },
-            setOf(this)
-        )
+        defer {
+            run {
+                val height = Units.Meters.of(tuningHeight.get())
+                setpointValue = height
+                io.setHeight(height)
+            }
+                .withName("Elevator/Tuning")
+        }
 
     fun setVoltage(voltage: Voltage): Command =
         startEnd(
--- a/src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
+++ b/src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
@@ -81,7 +81,7 @@ class Elevator(private val io: ElevatorIO) : SubsystemBase() {
             .withName("Elevator/setVoltage")
 
     fun manualControl(percentOutput: DoubleSupplier): Command =
-        runOnce { io.setVoltage(Units.Volts.of(percentOutput.asDouble * 12.0 + 0.37)) }
+        run { io.setVoltage(Units.Volts.of(percentOutput.asDouble * 12.0)) }
 
     fun stop(): Command = setVoltage(Units.Volts.zero()).withName("Elevator/stop")
 
--- a/src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
+++ b/src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
@@ -12,9 +12,11 @@ import org.littletonrobotics.junction.Logger
 import org.littletonrobotics.junction.mechanism.LoggedMechanism2d
 import org.littletonrobotics.junction.mechanism.LoggedMechanismLigament2d
 import org.littletonrobotics.junction.networktables.LoggedNetworkNumber
+import java.util.function.DoubleSupplier
 
 class Elevator(private val io: ElevatorIO) : SubsystemBase() {
-    @AutoLogOutput private val mechanism = LoggedMechanism2d(3.0, 3.0)
+    @AutoLogOutput
+    private val mechanism = LoggedMechanism2d(3.0, 3.0)
     private val root = mechanism.getRoot("Elevator", 2.0, 0.0)
     private val elevatorLigament =
         root.append(LoggedMechanismLigament2d("ElevatorLigament", 5.0, 90.0))
@@ -22,7 +24,8 @@ class Elevator(private val io: ElevatorIO) : SubsystemBase() {
     @AutoLogOutput
     private var setpointValue: Distance = Units.Millimeters.of(0.0)
 
-    @AutoLogOutput private var setpointName: Positions = Positions.ZERO
+    @AutoLogOutput
+    private var setpointName: Positions = Positions.ZERO
 
     @AutoLogOutput
     private val isStuck = Trigger {
@@ -51,10 +54,13 @@ class Elevator(private val io: ElevatorIO) : SubsystemBase() {
     fun l4(): Command = setHeight(Positions.L4).withName("Elevator/L4")
     fun l2Algae(): Command =
         setHeight(Positions.L2_ALGAE).withName("Elevator/L2 Algae")
+
     fun l3Algae(): Command =
         setHeight(Positions.L3_ALGAE).withName("Elevator/L3 Algae")
+
     fun feeder(): Command =
         setHeight(Positions.FEEDER).withName("Elevator/Feeder")
+
     fun zero(): Command =
         setHeight(Positions.ZERO).withName("Elevator/Move To Zero")
 
@@ -64,13 +70,14 @@ class Elevator(private val io: ElevatorIO) : SubsystemBase() {
             setpointValue = height
             io.setHeight(height)
         }
-            .withName("Elevator/Tuning")}, setOf(this))
+            .withName("Elevator/Tuning")
+    }, setOf(this))
 
     fun setVoltage(voltage: Voltage): Command =
         startEnd(
-                { io.setVoltage(voltage) },
-                { io.setVoltage(Units.Volts.zero()) }
-            )
+            { io.setVoltage(voltage) },
+            { io.setVoltage(Units.Volts.zero()) }
+        )
             .withName("Elevator/setVoltage")
 
     fun manualControl(percentOutput: DoubleSupplier): Command =
--- a/src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
+++ b/src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
@@ -80,7 +80,7 @@ class Elevator(private val io: ElevatorIO) : SubsystemBase() {
         )
             .withName("Elevator/setVoltage")
 
-    fun manualControl(percentOutput: DoubleSupplier): Command =
+    fun powerControl(percentOutput: DoubleSupplier): Command =
         run { io.setVoltage(Units.Volts.of(percentOutput.asDouble * 12.0 + VOLTAGE_CONTROL_KG)) }
 
     fun stop(): Command = setVoltage(Units.Volts.zero()).withName("Elevator/stop")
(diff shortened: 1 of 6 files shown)


src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
```diff
@@ -32,16 +36,16 @@ class Elevator(private val io: ElevatorIO) : SubsystemBase() {
...
-        runOnce {
-                setpointValue = height.value
-                setpointName = height
-                io.setHeight(height.value)
-            }
+        run {
+            setpointValue = height.value
+            setpointName = height
+            io.setHeight(height.value)
+        }
```
katzuv, line 48: Why is that needed?

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
```diff
@@ -50,29 +54,45 @@ class Elevator(private val io: ElevatorIO) : SubsystemBase() {
...
-    fun tuningPosition(): Command =
-        run { io.setHeight(Units.Meters.of(tuningHeight.get())) }
+    fun tuningPosition(): Command = Commands.defer({
+        runOnce {
+            val height = Units.Meters.of(tuningHeight.get())
+            setpointValue = height
+            io.setHeight(height)
+        }
             .withName("Elevator/Tuning")
+    }, setOf(this))
```
katzuv, line 74: Why not make this a `run()` command instead? That way you won't need to trigger the command when the tuned height is changed.

src/main/kotlin/frc/robot/subsystems/elevator/ElevatorConstants.kt
```diff
@@ -12,27 +12,27 @@ import frc.robot.lib.Gains
...
 val SPROCKET_RADIUS: Distance = Units.Millimeters.of(36.4 / 2)
 private val ROTATIONS_TO_CENTIMETERS_RATIO =
     GEAR_RATIO *
-        FIRST_STAGE_RATIO *
+        SECOND_STAGE_RATIO *
         (SPROCKET_RADIUS.`in`(Units.Centimeter) * 2 * PI)
 val ROTATIONS_TO_CENTIMETER: Measure<out PerUnit<DistanceUnit, AngleUnit>> =
     Units.Centimeters.per(Units.Rotations).of(ROTATIONS_TO_CENTIMETERS_RATIO)
 val CENTIMETERS_TO_ROTATIONS: Measure<out PerUnit<AngleUnit, DistanceUnit>> =
     Units.Rotations.per(Units.Centimeter).of(1 / ROTATIONS_TO_CENTIMETERS_RATIO)
```
katzuv, line 32: Wasn't that moved to `lib`?

src/main/kotlin/frc/robot/subsystems/elevator/ElevatorIOReal.kt
```diff
@@ -64,18 +62,10 @@ class ElevatorIOReal : ElevatorIO {
...
                 Feedback = FeedbackConfigs().apply { RotorToSensorRatio = 1.0 }
                 Slot0 = slot0Configs
-                HardwareLimitSwitch =
-                    HardwareLimitSwitchConfigs().apply {
-                        ReverseLimitType = ReverseLimitTypeValue.NormallyOpen
-                        ReverseLimitEnable = true
-                        ReverseLimitSource =
-                            ReverseLimitSourceValue.RemoteTalonFX
-                        ReverseLimitRemoteSensorID = EXTENDER_MOTOR_ID
-                    }
```
katzuv, line 78: This should be reverted.

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
```diff
@@ -50,29 +54,45 @@ class Elevator(private val io: ElevatorIO) : SubsystemBase() {
...
-        setVoltage(RESET_VOLTAGE)
-            .until(isStuck)
-            .andThen(runOnce(io::reset))
-            .withName("Elevator/reset")
+    fun manualControl(percentOutput: DoubleSupplier): Command =
+        runOnce { io.setVoltage(Units.Volts.of(percentOutput.asDouble * 12.0 + 0.37)) }
+
+    fun stop(): Command = setVoltage(Units.Volts.zero()).withName("Elevator/stop")
+
+    fun reset(resetTrigger: Trigger): Command =
```
katzuv, line 88: The trigger name is not clear. It should be emphasized that the elevator's current has reached the resetting threshold.

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
```diff
@@ -50,29 +54,45 @@ class Elevator(private val io: ElevatorIO) : SubsystemBase() {
...
+        )
             .withName("Elevator/setVoltage")
 
-    fun reset(): Command =
-        setVoltage(RESET_VOLTAGE)
-            .until(isStuck)
-            .andThen(runOnce(io::reset))
-            .withName("Elevator/reset")
+    fun manualControl(percentOutput: DoubleSupplier): Command =
+        runOnce { io.setVoltage(Units.Volts.of(percentOutput.asDouble * 12.0 + 0.37)) }
```
katzuv, line 84: 1. 0.37 is a magic number. Why is that different than the kG in the constants file?
2. It's better to just use `DutyCycleOut` in this case.
3. Why not make the joystick change the height position instead?
4. There should be some protection that prevents 100% power.
5. Why `runOnce()`? Don't we want to poll the supplier continuously?
6. The name of the command should change:
   1. Remove "manual" -- this will work with any supplier.
   2. Indicate what exactly it controls (power, or height if 3 is implemented).
7. This command should be combined with `setVoltage()` (or at least one should use the other).

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
```diff
@@ -50,29 +54,45 @@ class Elevator(private val io: ElevatorIO) : SubsystemBase() {
...
+
+    fun stop(): Command = setVoltage(Units.Volts.zero()).withName("Elevator/stop")
```
katzuv, line 86: What's the use case?

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
Emma03L: It's no longer current based. The current threshold changes when resetting from different heights.

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
katzuv: Where is it configured?

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
katzuv: Why is/was the defer needed?

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
katzuv: Doesn't this command work by moving the elevator all the way down? How does that changed when the initial height chnages?

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
Emma03L: 1. ab4aab30214705c0df8517628040cb7210f5b323
2. Because `tuningHeight` is only updated when the command is created

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
Emma03L: We didn't want to commit the robot container changes

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
Emma03L: Instead of using `setVoltage(Units.Voltz.zero)` we can use `stop()`. It's shorter and more readable

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
Emma03L: 1. That was accidentally commited from testing. oops
2. That would require an io function that is really unnecessary.
3. This is meant as a manual control/override. A way to control the elevator reliably even if there's a problem with the position control.
4. 👍
5. 👍
6. 👍
7. It does use `setVoltage()`. It's a wrapper command for a specific use case.

rakrakon: 1. 9212af0177c0468a4db24ee9ddeff7fa79c4b377
2. Why? We don't want different speeds based on the battery voltage. Also this might hurt the `reset` function if we need to reset towards the ends of the game.
3. If we do a sudden movement we can accidently do a drastic change in the position resulting in slamming the elevator in either direction. Also in my experience this type of control is very funky and needs some calibration, since this is working great i don't see why we should switch.

Emma03L: 5. 36a287ced0b45da1ed8ebf73c1d6b94ef8416abe
6. 58859d7c4c093c18f3f87e9a696e587790be7724

src/main/kotlin/frc/robot/subsystems/elevator/ElevatorConstants.kt
Emma03L: 58859d7c4c093c18f3f87e9a696e587790be7724

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
katzuv: Yes but, when would we need to "stop" the elevator that way?

rakrakon: In the reset command

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
katzuv: The reset process detects when there's a current spike which indicates the motor stalls, no? How does that depend on the initial height?

Emma03L: The current spike is different depending on the initial height

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
katzuv: Ok, but what is the default command?

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
katzuv: If that's the only use case I'd just that write that directly in the reset command. Anyway, it should private

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
katzuv: 2. OK
3. Then don't do sudden movements, that's what (4) is for. I don't see why that'd be funky as long as the setpoint is moved slowly enough.
4. Soft limits are like bang-bang control: they don't smooth the movement near the limits. So if you slam the elevator with 100%, yes the motor will stop when the limit is reached, but it won't be instant. You should add some damping near the limits.

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
katzuv: OK. Note that you can use the subsystem's defer

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
Emma03L: 3. 

> This is meant as a manual control/override. A way to control the elevator reliably even if there's a problem with the position control.

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
Emma03L: There could be other use cases in the future. There is no reason to remove this or make it private

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
katzuv: Sorry, missed that.

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
katzuv: If we'll need it we can bring it back. Delete it in one commit, give it an indicative message, and then you can revert it easily.
We've don't have "stop" commands in (most?) other subsystems. See [YAGNI](https://en.m.wikipedia.org/wiki/You_aren%27t_gonna_need_it) (You Ain't Gonna Need It).

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
rakrakon: We are resetting the elevator like that intake last year. We are not using current since we found it to be inconsistent.

katzuv: And… how's that done? If you're not using current, what's the `isStuck` trigger for?

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
Emma03L: ed6448cd4105bfc3828460bbe33370b26a4ad909

katzuv: I'm pretty sure you need to use `withName()` on the `defer()`, not `run()`.

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
rakrakon: It can be removed

katzuv: Barak, can you please give me full answers? I don't know how the intake was reset last year. Who remembers? Just answer, it will make these threads much shorter and quicker to resolve :)
Also, if current isn't used (although that's a popular method across FRC teams and in the industry as well, so I'm sure you can make it work), what current spike does Emma refer to?

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
Emma03L: The logic works, we checked it

katzuv: Where did I say it doesn't? I'm genuinely trying to understand how that works

Emma03L: The default command always runs, when it's interrupted by another command requiring the same subsystem it stops until the other command is finished

src/main/kotlin/frc/robot/subsystems/elevator/ElevatorIOReal.kt
Emma03L: 29313e2dccb8bb51708bfc33b33f96c9da764e3b

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
Emma03L: When going down from a higher starting point the system will have more time to accelerate, because it uses power control, so it will hit the bottom harder thus generating more current.

src/main/kotlin/frc/robot/subsystems/elevator/Elevator.kt
Emma03L: I don't think it's that bad that there's a command that isn't used a lot. It's like the sysid commands in the swerve, they aren't usually used but we don't want to delete and restore them every time we want to use them

PR #159, opened 2025-02-06
Wrist calibrations

Emma03L:


src/main/kotlin/frc/robot/subsystems/wrist/WristConstants.kt
```diff
@@ -7,11 +7,11 @@ import edu.wpi.first.units.measure.MomentOfInertia
...
 val MIN_ANGLE: Angle = Units.Rotations.of(0.0)
-const val GEAR_RATIO = 1.0 // TODO: Calibrate
-const val ROTOR_TO_SENSOR = 1.0
-const val SENSOR_TO_MECHANISM = 1.0
+const val ROTOR_TO_SENSOR = 1 / ((1.0 / 25.0) * (40.0 / 40.0))
+const val SENSOR_TO_MECHANISM = 1 / (16.0 / 42.0)
+const val GEAR_RATIO = 1 / ((1.0 / 25.0) * (40.0 / 40.0) * (16.0 / 42.0))
 val MOMENT_OF_INERTIA: MomentOfInertia = Units.KilogramSquareMeters.of(0.001)
-const val ENCODER_OFFSET = 0.0
+val ENCODER_OFFSET: Angle = Units.Radians.of(0.4095728703654662)
```
katzuv, line 14: Why radians?

src/main/kotlin/frc/robot/subsystems/wrist/WristIOReal.kt
```diff
@@ -72,13 +84,6 @@ class WristIOReal : WristIO {
                     }
             }
         )
-
-        val encoderConfig =
-            CANcoderConfiguration().apply {
```
katzuv, line 77: Why was that moved?

katzuv (changes requested):

src/main/kotlin/frc/robot/subsystems/wrist/WristConstants.kt
Emma03L: Those are the default units of angle measures in advantage scope

src/main/kotlin/frc/robot/subsystems/wrist/WristIOReal.kt
Emma03L: We thought it would solve a problem, it wasn't related but we already committed it and it really doesn't matter.

src/main/kotlin/frc/robot/subsystems/wrist/WristConstants.kt
katzuv: And what prevents you from using the unit conversion feature?
Again, use the best unit you can. That is degrees in this case. If that'll need to change quickly in the future, no one will want to start converting units under pressure.

src/main/kotlin/frc/robot/subsystems/wrist/WristIOReal.kt
katzuv: Would the code work after this change? Yes. 
Would it be more elegant and readable that all configurations were in the same place? Also yes. And you should strive to write elegant code.

katzuv: Would the code work after this change? Yes. 
Would it be more elegant and readable that all configurations were in the same place? Also yes. And you should strive to write elegant code.

src/main/kotlin/frc/robot/subsystems/wrist/WristConstants.kt
Emma03L: Using degrees will cause this exact inconvenience and pressure you're describing. The point of measures is that we can assign the value in the most convenient unit and use it in the most convenient unit. In this situation the most convenient unit to assign with is radians, because when you open advantage scope and select the no offset encoder position you will see it in radians, to change it to degrees in advantage scope requires an extra step which creates an inconvenience.

src/main/kotlin/frc/robot/subsystems/wrist/WristIOReal.kt
Emma03L: They are in the same place. The encoder is now configured above the motor. Their order is an arbitrary decision we made, it doesn't change the code's elegance.

src/main/kotlin/frc/robot/subsystems/wrist/WristConstants.kt
rakrakon: I don't see how setting the solution with "our eyes" is better than putting in the value. That just creates unnecessary inaccuracies. The best thing to do here is make is as convenient as possible to update the offset accurately. We don't want to deal with conversion or anything like that because there's simply no need or it and it will just be an over complication.

src/main/kotlin/frc/robot/subsystems/wrist/WristIOReal.kt
Emma03L: 65e0e25cfe747ac414e9e0b59b819d240372c85a

src/main/kotlin/frc/robot/subsystems/wrist/WristConstants.kt
katzuv: My point is you'll need AdvantageScope to change the value. If you do it with degrees and just want to slightly change it, it's easier to do so with degrees because they are more convenient for humans than radians and you won't necessarily need to use any external software (AdvantageScope/Tuner X). You can resolve the conversation if you disagree, it's not a critical issue.

PR #167, opened 2025-02-09
New climb

Emma03L:


src/main/kotlin/frc/robot/subsystems/climber/ClimberConstants.kt
```diff
@@ -49,23 +24,10 @@ var MOTOR_CONFIG =
             StatorCurrentLimit = 160.0
             SupplyCurrentLimit = 80.0
         }
-        Feedback.apply {
-            RotorToSensorRatio = ROTOR_TO_SENSOR
-            SensorToMechanismRatio = SENSOR_TO_MECHANISM
-            FeedbackRemoteSensorID = CANCODER_ID
-            FeedbackSensorSource = FeedbackSensorSourceValue.FusedCANcoder
-        }
         SoftwareLimitSwitch.apply {
```
katzuv, line 27: If there is no CANcoder now, I don't know how reliable the internal encoder will be, so perhaps the software limits should be removed.
Why, was the CANcoder removed? It will help us make the climb faster, which is important for a last-second climb (and RP).

katzuv (commented):
I think the subsystem was simplified too much. IMO it's very important to have a very fast climber. If we can climb (reliably) in a few seconds it will guarantee us a PR, and the "fastness" will increase the chance we climb even if we got stuck somewhere in the field, the driver took a few more seconds to align to the cage, etc.
I would:
- Bring back the second motor: if there is enough weight, the climber will be faster and the center of gravity will be lower.
- Bring back the stopper: this will ensure we get the climb RP even if we climbed very little. However, we need to check how much height the robot loses in 3 seconds, at a very low angle (there the moment of gravity is the largest).

src/main/kotlin/frc/robot/subsystems/climber/ClimberConstants.kt
Emma03L: I did remove the soft limits but didn't commit it because we're not sure if there will be a cancoder

PR #168, opened 2025-02-09
Make button bindings more concise

katzuv:
And other small changes.
- Refactor button bindings to be more concise
- Remove unneeded PathPlanner command setting function
- Remove unused variable
- Remove unneeded requirement
- Remove unneeded variables for subsystem
- Remove redundant newline
- Combine `visualizer` instantiation with declaration
- Add braces to `if` statement
- Replace variables with imports
- Fix `LoggedPowerDistribution` module type
- IntelliJ IDEA Reformat + Spotless

### DIFF
--- a/src/main/kotlin/frc/robot/RobotContainer.kt
+++ b/src/main/kotlin/frc/robot/RobotContainer.kt
@@ -86,7 +86,7 @@ object RobotContainer {
         driverController.apply {
             create()
                 .onTrue(
-                    Commands.runOnce(swerveDrive::resetGyro, swerveDrive)
+                    Commands.runOnce(swerveDrive::resetGyro)
                         .ignoringDisable(true)
                 )
             (listOf(
--- a/src/main/kotlin/frc/robot/RobotContainer.kt
+++ b/src/main/kotlin/frc/robot/RobotContainer.kt
@@ -32,8 +32,9 @@ object RobotContainer {
         configureButtonBindings()
         configureDefaultCommands()
 
-        if (CURRENT_MODE == Mode.SIM)
+        if (CURRENT_MODE == Mode.SIM) {
             SimulatedArena.getInstance().resetFieldForAuto()
+        }
 
         enableAutoLogOutputFor(this)
     }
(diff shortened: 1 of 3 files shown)


src/main/kotlin/frc/robot/RobotContainer.kt
```diff
@@ -87,44 +69,44 @@ object RobotContainer {
...
-        operatorController
-            .start()
-            .onTrue(feeder(operatorController.start().negate()))
-        operatorController
-            .povDown()
-            .onTrue(elevator.reset(operatorController.povDown().negate()))
-        operatorController
-            .povUp()
-            .onTrue(extender.reset(operatorController.povUp().negate()))
+        driverController.apply {
```
Emma03L, line 72: I think this is less readable and makes it harder to switch buttons between controllers

Emma03L (changes requested):
Looks great! But a bit unreadable

src/main/kotlin/frc/robot/RobotContainer.kt
katzuv: Perhaps it's for the better, otherwise there's more change for errors. The biggest improvement here is that you don't need to write the button twice (the negate part is done for you)

Emma03L: I like not having to write the negate but I still think this is less readable, maybe an extension function could be good.

katzuv: What would that exentsion function look like?
What exactly about the code looks unreadable?
@rakrakon, what is your opinion?

rakrakon: First of all I agree with Emma about the readability. As for the code duplication I think our best option is to do something like this:
'driverController.circle().run {
    onTrue(l3(negate()))
}'

Emma03L: > What would that exentsion function look like?

Something like this:
```  
private fun Trigger.startEnd(command: (endTrigger: Trigger) ->Command) =
    this.onTrue(command(this.negate()))
```
Not sure about the name.

katzuv: Ok

katzuv (commented):
Looks great! As long as you tested it and it works, you can merge.

PR #172, opened 2025-02-12
Swerve fix

Emma03L:


src/main/kotlin/frc/robot/subsystems/drive/Drive.java
```diff
@@ -434,9 +434,9 @@ public Pose2d getPose() {
         return poseEstimator.getEstimatedPosition();
     }
 
-    /** Returns the current gyro rotation. */
+    /** Returns the current gyro rotation or the estimated rotation if the gyro disconnects. */
```
katzuv, line 437: When do you need to get the gyro's measurement directly? Why not always use the estimated pose's rotation?
Not very well versed with other parts of the code so genuine question

Emma03L: I don't think we should trust the PE rotation all of the time

PR #174, opened 2025-02-13
Parse choreo poses

Emma03L:


src/main/kotlin/frc/robot/Constants.kt
```diff
@@ -13,6 +21,49 @@ private val SPEAKER_POSE_BLUE = Translation2d(0.0, 5.5479442)
...
+                ?: return emptyMap()
+
+        return jsonPoses.mapValues { (key, pose) ->
+            Pose2d(
+                pose.jsonObject["x"]
+                    ?.jsonObject
+                    ?.get("val")
+                    ?.jsonPrimitive
+                    ?.doubleOrNull
+                    ?: 0.0,
```
katzuv, line 48: I think this should be extracted into a function, it's used three times and the only thing that changes is the field name

src/main/kotlin/frc/robot/Constants.kt
```diff
@@ -13,6 +21,49 @@ private val SPEAKER_POSE_BLUE = Translation2d(0.0, 5.5479442)
 val SPEAKER_POSE: Translation2d
     get() = SPEAKER_POSE_BLUE.flipIfNeeded()
 
+val choreoPoses: Map<String, Pose2d>
+    get() {
```
katzuv, line 25: I don't know how expensive the parsing is. Why not parse it once on robot startup?

katzuv (changes requested):

src/main/kotlin/frc/robot/Constants.kt
Emma03L: 8ff98e74da6358997decd5416bd8b7e4a07db1a4

src/main/kotlin/frc/robot/Constants.kt
Emma03L: e45b6d3701b02c1a45197f4adda0719cbcd94f63

src/main/kotlin/frc/robot/Constants.kt
katzuv: I think it's better to throw an exception:
1. I'd rather find out before the match we have a problem with auto alignment (and maybe with auto period as well).
2. This will crash the first time w euse the auto alignment mid-match, because the code will try to access a non-existent field in the map.

src/main/kotlin/frc/robot/Constants.kt
```diff
@@ -13,6 +22,36 @@ private val SPEAKER_POSE_BLUE = Translation2d(0.0, 5.5479442)
...
+    return jsonPoses.mapValues { (key, pose) ->
+        Pose2d(
+            getValueFromJson(pose, "x"),
+            getValueFromJson(pose, "y"),
+            Rotation2d(getValueFromJson(pose, "heading"))
+            )
+    }
+}
+
+val choreoPoses = parseChoreoPoses()
```
katzuv, line 53:
```suggestion
val ALIGNMENT_POSES = parseChoreoPoses()
```
1. It's a constant. 
2. The fact it came from Choreo is an implementation detail.

Also, perhaps all this code should be moved to an autos directory? I think it will be convenient to have an autos directory, with sub-files or sub-directories for the auto period and for auto alignment.

katzuv (changes requested):

src/main/kotlin/frc/robot/Constants.kt
Emma03L: 969ac656be12ef2d852b143a00f3bcb5f6b92760

src/main/kotlin/frc/robot/Constants.kt
Emma03L: d74c898240c5090af47fa0387b7799900b9b1e7f

katzuv: Great. What's your opinion on the second thing?

Emma03L: Oh I missed it, yeah I think it's a good idea' I'll do it

Emma03L: 96d34a7beaf25a17f5a765d9c5d13de174831926

katzuv (approved):

PR #179, opened 2025-02-14
Optional maple sim

Emma03L:


src/main/kotlin/frc/robot/Constants.kt
```diff
@@ -22,6 +22,10 @@ val CURRENT_MODE: Mode
                 Mode.SIM
             }
         }
+
+val USE_MAPLE_SIM: Boolean
+    get() = System.getenv()["isMapleSim"] == "true"
```
katzuv, line 27: Copying from line 32:

```suggestion
    get() = System.getenv("isMapleSim") == "true"
```

katzuv (changes requested):

src/main/kotlin/frc/robot/Constants.kt
Emma03L: In that case `System.getenv()["isReplay"] == "true"` should also change for consistency.
dd7a6e883032aecd8def342c9c701009cd293219

katzuv (approved):
As always, I think it's best to solve issues than write patches, but let's hope this helps us fix the auto problems

src/main/kotlin/frc/robot/Constants.kt
katzuv: Awesome

Emma03L:
> As always, I think it's best to solve issues than write patches, but let's hope this helps us fix the auto problems

The poses not being synchronized was already dealt with, this is just makes it optional because I think it's a good idea.

PR #180, opened 2025-02-15
Fix power distribution

Emma03L:


src/main/kotlin/frc/robot/Robot.kt
```diff
@@ -68,8 +68,8 @@ object Robot : LoggedRobot() {
         when (CURRENT_MODE) {
             REAL -> {
                 LoggedPowerDistribution.getInstance(
-                    0,
-                    PowerDistribution.ModuleType.kCTRE
+                    1,
+                    PowerDistribution.ModuleType.kRev
                 )
```
katzuv, line 73: Looks like we use the default ID, so should be just: 
```kt
LoggedPowerDistribution()
```
https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/wpilibj/PowerDistribution.html#%3Cinit%3E()

katzuv (changes requested):

src/main/kotlin/frc/robot/Robot.kt
Emma03L: I tried the default constructor and it still didn't log

katzuv: And now it does?
If so, we should open an issue to AdvantageKit or WPILib. Anyway, if it works you can merge.

katzuv (approved):

PR #186, opened 2025-02-22
Add 3D Alignment

Emma03L:


src/main/kotlin/frc/robot/RobotContainer.kt
```diff
@@ -41,6 +47,16 @@ object RobotContainer {
...
 
+    private val Reef3: Pose2d =
+        getPose2d(3.769, 5.416, Rotation2d.fromDegrees(-60.0))
+    private val Reef4: Pose2d = getPose2d(2.93, 4.030)
+    private val Reef5: Pose2d =
+        getPose2d(3.719, 2.644, Rotation2d.fromDegrees(60.0))
+    private val Reef6: Pose2d =
+        getPose2d(5.255, 2.624, Rotation2d.fromDegrees(120.0))
+
+    private val selectedPose: Pose2d = Reef6
```
katzuv, line 58: shouldn't be here

src/main/kotlin/frc/robot/RobotContainer.kt
```diff
@@ -94,6 +110,20 @@ object RobotContainer {
             .onTrue(outtakeAlgae(driverController.L1().negate()))
         driverController.R2().whileTrue(gripper.intake())
         driverController.L2().whileTrue(gripper.outtake())
+        driverController
+            .povLeft()
+            .whileTrue(
+                pathFindToPose(selectedPose)
+                    .andThen(alignToPose(swerveDrive, true, l4()))
+            )
+            .onFalse(l4(Trigger { true }))
```
katzuv, line 119:
```suggestion
            .onFalse(l4({ true }))
```
I think this will work because `Trigger` inherits `BooleanSupplier`.

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
```diff
@@ -8,76 +8,66 @@ import edu.wpi.first.math.kinematics.ChassisSpeeds
...
-        PIDController(
-            ALIGNMENT_ROTATION_GAINS.kP,
-            ALIGNMENT_ROTATION_GAINS.kI,
-            ALIGNMENT_ROTATION_GAINS.kD,
-        )
-    rotationController.enableContinuousInput(-Math.PI, Math.PI)
-
-    radialController.setTolerance(LINEAR_ALIGNMENT_TOLERANCE.`in`(Units.Meters))
+        ALIGNMENT_ROTATION_GAINS.run { PIDController(kP, kI, kD) }
+    rotationController.setpoint = Rotation2d.fromDegrees(155.0).radians
```
katzuv, line 24: Magic number

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
```diff
@@ -8,76 +8,66 @@ import edu.wpi.first.math.kinematics.ChassisSpeeds
...
     )
 
-    return Commands.run(
-            {
-                val robotPose = robotPoseSupplier.get()
-                val targetPose = targetPoseSupplier.get()
+    val xController = ALIGNMENT_X_GAINS.run { PIDController(kP, kI, kD) }
+    xController.setpoint = if (isLeft) -ALIGNED_TX_LEFT else -ALIGNED_TX_RIGHT
+    xController.setTolerance(LINEAR_ALIGNMENT_TOLERANCE.`in`(Units.Meters))
+    val xError = { -vision.getTranslationToBestTarget(1).y }
```
katzuv, line 32: X error or Y error? Because you take they component of the translation. 
Also, 1 is a magic number

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
```diff
@@ -8,76 +8,66 @@ import edu.wpi.first.math.kinematics.ChassisSpeeds
...
+                        0.0
                     )
                 )
-            },
-            swerveDrive
+            }
         )
-        .until { radialController.atSetpoint() }
-        .andThen(Commands.runOnce({ drive.runVelocity(ChassisSpeeds()) }))
+        .alongWith(Commands.run({ Logger.recordOutput("XError", xError) }))
```
katzuv, line 72: Add outputs for Y and heading as well, and place then an under a subtable

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
```diff
@@ -8,76 +8,66 @@ import edu.wpi.first.math.kinematics.ChassisSpeeds
...
-                            robotPose.rotation.radians,
-                            targetPose.rotation.radians
+    val yController = ALIGNMENT_Y_GAINS.run { PIDController(kP, kI, kD) }
+    yController.setTolerance(LINEAR_ALIGNMENT_TOLERANCE.`in`(Units.Meters))
+    yController.setpoint = 0.0
+    val yError = { -vision.getTranslationToBestTarget(1).x }
+    return Commands.sequence(
+            drive
+                .run {
+                    drive.runVelocity(
```
katzuv, line 41: Add another command, it's weird to write drive twice

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
```diff
@@ -8,76 +8,66 @@ import edu.wpi.first.math.kinematics.ChassisSpeeds
...
+            WaitCommand(0.3),
+            drive.run {
                 drive.runVelocity(
-                    ChassisSpeeds.fromFieldRelativeSpeeds(
-                        targetSpeeds,
-                        if (IS_RED && CURRENT_MODE == Mode.REAL)
-                            drive.rotation + Rotation2d.k180deg
-                        else drive.rotation
+                    ChassisSpeeds(
+                        ALIGNMENT_Y_VELOCITY.`in`(Units.MetersPerSecond),
```
katzuv, line 65: What does that do? the name of the constant is also not clear

katzuv (changes requested):

src/main/kotlin/frc/robot/RobotContainer.kt
Emma03L: We checked, it didn't

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
Emma03L: af4b136ff047f41dad7d053bb7e5240f03f5ece9

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
Emma03L: f10a17fdf1674f476839c22632e25df4e336be91

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
Emma03L: Because the camera doesn't always see the tag when aligned we just drive forward

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
Emma03L: 6b79c028975808640bbb0805df78426f95e8c698

Emma03L: 5bcbcc13b40f8177d4b2375401973e5f5fa1646d Removed unused controller

src/main/kotlin/frc/robot/RobotContainer.kt
katzuv: Ok but you can change the type that `l4` takes

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
katzuv: This doesn't make much sense:
https://github.com/Galaxia5987/robot-2025/blob/af4b136ff047f41dad7d053bb7e5240f03f5ece9/src/main/kotlin/frc/robot/autonomous/AutoCommands.kt#L30
And 1 is still a magic number.

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
katzuv: > [6b79c02](https://github.com/Galaxia5987/robot-2025/commit/6b79c028975808640bbb0805df78426f95e8c698)

I'd put it under an auto-alignment subtable, it's not really related to vision.

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
katzuv: OK. Is there a reason to not put any timeout on this command?

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
```diff
@@ -8,76 +8,64 @@ import edu.wpi.first.math.kinematics.ChassisSpeeds
...
+                .until(
+                    Trigger {
+                            yController.atSetpoint() &&
+                                rotationController.atSetpoint()
+                        }
+                        .debounce(0.15)
+                ),
+            drive
+                .runOnce { drive.setAngle(Rotation2d.kZero) }
+                .alongWith(scoreCommand),
```
katzuv, line 56: I'd raise the elevator a bit before. A nice way to do that is to create the `PIDController`s outside this function (but make them private). Then make a trigger that checks when the controller's error is smaller than some value, and this trigger's `onTrue()` will schedule the `scoreCommand`. Make sure this doesn't make the elevator think it has multiple commands (in other words, make sure the trigger schedules only one score command, multiple `onTrue()` calls might schedule multiple commands).

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
Emma03L: Yes we checked

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
Emma03L: I don't think this is necessary

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
Emma03L: 7a685cbf3d0710fef2a1bbb23328963490263858

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
katzuv: Shortening cycles is beneficial. We can open an issue about that and add that in another PR (possibly after DIS2).

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
Emma03L: We want it to stop when the driver sees that the robot is stuck, not necessarily after a fixed amount of time

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
katzuv: A few things:
- Now I'm not sure this will work, `yError` is a function, not a double.
- I believe you can remove the `( )`:
    ```kt
    Commands.run { Logger.recordOutput("Auto Alignment/YError", yError) }
    ```
- Another alternative: create the error getters in the file itself (and private), then use `@AutoLogOutput`.
- I recommend adding another output that says whether we are now using auto alignment. We can also use this one to trigger a special LEDs pattern.

src/main/kotlin/frc/robot/RobotContainer.kt
Emma03L: I don;t think it's necessary

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
Emma03L: dee194beea6b39dc3fbe88a7c87b1ab21c1e9d26

Emma03L: 25b340e00a05cc5342d830a36af803d47f1cfc0f

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
Emma03L: f71f75e17fdedb760e80f415e415443bc99a4b4b

src/main/kotlin/frc/robot/RobotContainer.kt
katzuv: It's cleaner and if you can use a "broader" type -- you should. This will make your code more flexible.

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
katzuv: Awesome

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
katzuv: The timeout can be 10 or 20 seconds. So if we're in the shop and we're checking something else we won't forget the robot is actually driving, which might lead to safety issues.

.kotlin/sessions/kotlin-compiler-8758221280121607423.salive
katzuv, line 1: Remove

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
katzuv: Well, now you supply a constant `ChassisSpeeds` object. Perhaps in hindsight, it's cleaner to revert to the previous version, but you can also give `driveCommand` a supplier of `ChassisSpeeds`.

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
Emma03L: We checked, the log works

src/main/kotlin/frc/robot/RobotContainer.kt
Emma03L: I don't want to mess with stuff that works. I know it shouldn't mess anything but it's not critical enough to do it 2 days before a competition

src/main/kotlin/frc/robot/RobotContainer.kt
Emma03L: Will be moved when attaching the operator buttons (different branch)

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
katzuv: These outputs will help us. I'm sure that by now you could have already made the changes. It shouldn't take more than two minutes.

.kotlin/sessions/kotlin-compiler-8758221280121607423.salive
katzuv, line 1: Remove

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
```diff
@@ -8,76 +8,62 @@ import edu.wpi.first.math.kinematics.ChassisSpeeds
...
-                            drive.rotation + Rotation2d.k180deg
-                        else drive.rotation
-                    )
+    return Commands.sequence(
+        DriveCommands.driveCommand(
+            drive, ChassisSpeeds(
+                0.0,
+                yController.calculate(yError.invoke()),
+                -rotationController.calculate(
+                    vision.getYawToTarget(VisionConstants.frontCameraIndex).get().radians
```
katzuv, line 42: `vision.getYawToTarget(VisionConstants.frontCameraIndex).get().radians` should be extracted to a variable like `yError`. If you don't want to fix that now, open an issue.

katzuv (changes requested):

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
Emma03L: 14c9776ce4f9ac0c8105276af9f3cbd8ffd3b83b

Emma03L: 61460a2a9e2b5fb067fb90a52841c193c0f69aba

.kotlin/sessions/kotlin-compiler-8758221280121607423.salive
Emma03L: 17d0981a731923dbe9bb5651434ac358b253d667

Emma03L: 17d0981a731923dbe9bb5651434ac358b253d667

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
Emma03L: 6d9041dee53c75ef57eddd0a87c128c057901225

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
katzuv: Awesome!
But, you don't log an output, and why run `aligning = false` only if the command was interrupted? Why not just when it ends?

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
Emma03L: 4e5841f884dbf03b920a627f4d805590ae00906c

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
Emma03L: I'm pretty sure interrupted is also true when the command ends but if the command is interrupted then it won't be considered ended
827828e2365de62d3c7f224f1cf883bab60d2aed

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
katzuv: I'd argue that this change is bigger than changing triggers to `() -> Boolean`, but I won't argue with you making it 😜 
Anyway, an output for this is still missing

src/main/kotlin/frc/robot/autonomous/AutoCommands.kt
katzuv: No, a command is considered interrupted only if it was interrupted by another command. But it's considered ended whenever it ended. This is why `Command::end()` takes a boolean, which is whether the command was interrupted or not. You can also always check the source and see that is uses `finallyDo()`:
```java
public WrapperCommand handleInterrupt(Runnable handler) {
  requireNonNullParam(handler, "handler", "Command.handleInterrupt()");
  return finallyDo(
      interrupted -> {
        if (interrupted) {
          handler.run();
        }
      });
}
```
https://github.com/wpilibsuite/allwpilib/blob/f9307de04c55e437eaf6b9f8320819ba8f7db9bf/wpilibNewCommands/src/main/java/edu/wpi/first/wpilibj2/command/Command.java#L526-L534
Anyway, if you want this to run just when it ends, regardless of why it ended (which is what we need here).

The latest commit number you attached is not relevant.

katzuv: About the output:
1. I believe you don't need the function -- we don't use a function when logging `Mechanism2d`s, for example.
2. I don't know which subtable this will be logged under, but it's not going to be `Auto Alignment`. You should use the `key` parameter here (something like `Auto Alignment/is aligning`. Check the docs: https://docs.advantagekit.org/data-flow/recording-outputs/annotation-logging/

Emma03L: 6bdd3f06df9b5fc0e1290c0522a79c024f0fb59c

PR #199, opened 2025-03-04
Autonomous

Emma03L:


src/main/deploy/pathplanner/paths/6LS.path
katzuv, line 1: The rotation target is redundant, the robot doesn't seem to be close to hitting someting

src/main/kotlin/frc/robot/autonomous/AutonomousCommands.kt
```diff
@@ -0,0 +1,112 @@
...
+        Commands.runOnce({ selectedScorePose = { Reef5Left.flipIfNeeded() } }),
+        autoScoreL4()
+    )
+
+private fun S5R(): Command =
+    Commands.sequence(
+        AutoBuilder.followPath(PathPlannerPath.fromPathFile("S5R")),
+        Commands.runOnce({ selectedScorePose = { Reef5Right.flipIfNeeded() } }),
+        autoScoreL4()
+    )
```
katzuv, line 67: These commands are pretty identical. Extract the logic to a private function that takes parameters. Whether to reset the odometry or not should also be a parameter.

src/main/kotlin/frc/robot/autonomous/ProfiledPosePID.kt
```diff
@@ -118,7 +118,15 @@ fun getSpeed(botPose: Pose2d): () -> ChassisSpeeds {
             Logger.recordOutput("AutoAlignment/$key", value)
         }
 
-    Logger.recordOutput("AutoAlignment/AtGoal", atGoal)
+    mapOf(
+            "AtGoal" to atGoal.asBoolean,
+            "XAtSetpoint" to xController.atSetpoint(),
+            "YAtSetpoint" to yController.atSetpoint(),
+            "thetaAtSetpoint" to thetaController.atSetpoint()
```
katzuv, line 125: Shouldn't you use `atGoal()`?

katzuv (changes requested):
A few general suggestions:
- Use PathPlanner's Optimize Path feature -- for some paths it reduces the path runtime by half a second.
- Start the robot in auto as close to the reef as possible:
  - Try a few starting points (move it up and down on the starting line), hit Optimize Path, and see if it helps.
  - We don't need to place the robot such that its center is on the starting line, we only need that the "BUMPERS overlap their ROBOT STARTING LINE" (from the manual).
- You can also try different points on the feeder.
- Looks like we drive forward after driving a path to the feeder. If so, the [goal end state](https://pathplanner.dev/gui-editing-paths-and-autos.html#goal-end-state) should have the same velocity we use in the driving forward part.

Excited to see the robot score a lot of points in autonomous! 🤩

src/main/kotlin/frc/robot/subsystems/ReefCommands.kt
```diff
@@ -89,7 +89,7 @@ fun outtakeCoral(): Command =
                 ))
             .withTimeout(0.5),
         gripper.slowOuttake(true).withTimeout(0.1),
-        moveDefaultPosition().onlyIf(gripper.hasCoral.negate())
+        moveDefaultPosition().onlyIf(gripper.hasCoral.negate().and{!DriverStation.isAutonomous()})
```
katzuv, line 92: Use [`RobotModeTriggers.teleop()`](https://github.wpilib.org/allwpilib/docs/release/java/edu/wpi/first/wpilibj2/command/button/RobotModeTriggers.html#teleop()) instead:
```suggestion
            .and(RobotModeTriggers.teleop())
```
Why is that change needed?

src/main/kotlin/frc/robot/subsystems/ReefCommands.kt
```diff
@@ -88,13 +89,17 @@ fun outtakeCoral(): Command =
...
 
 // TODO: Add Coral Simulation
 
 fun moveDefaultPosition(): Command =
-    sequence(elevator.feeder(), waitUntil(elevator.atSetpoint), wrist.feeder())
+    sequence(
+            elevator.feeder(),
+            waitUntil(elevator.atSetpoint),
+            wrist.feeder()
+        )
```
katzuv, line 102:
```suggestion
    elevator.run { feeder().until(atSetpoint()) }.andThen(wrist.feeder())
```
As usual, this is clearer using infix notation... 😉
```kt
elevator.run { feeder() until atSetpoint() } andThen wrist.feeder()
```
Edit: I'm not sure if that'll work because of `Subsystem.run()`. An alternative is to use `let` or `with`, but the benefit will be smaller:
```suggestion
    elevator.let { it.feeder().until(it.atSetpoint()) }.andThen(wrist.feeder())
```
```suggestion
    with(elevator) { feeder().until(atSetpoint()) }.andThen(wrist.feeder())
```

katzuv (changes requested):

src/main/kotlin/frc/robot/subsystems/ReefCommands.kt
```diff
@@ -88,7 +89,10 @@ fun outtakeCoral(): Command =
                 ))
             .withTimeout(0.5),
         gripper.slowOuttake(true).withTimeout(0.1),
-        moveDefaultPosition().onlyIf(gripper.hasCoral.negate())
+        moveDefaultPosition()
+            .onlyIf(
+                gripper.hasCoral.negate().and { !DriverStation.isAutonomous() }
```
katzuv, line 94: Use `RobotModeTriggers.teleop()` instead.

katzuv (changes requested):

PR #202, opened 2025-03-06
swerve disable commands

yakla:


src/main/kotlin/frc/robot/RobotContainer.kt
```diff
@@ -177,6 +168,20 @@ object RobotContainer {
...
 
+    val disableTrigger: Trigger =
+        RobotModeTriggers.disabled()
+            .debounce(disableCommandDebounce)
+            .onTrue(
+                Commands.runOnce({
+                    swerveDrive.SetNeutralMode(NeutralModeValue.Coast)
+                })
+            )
+            .onFalse(
```
katzuv, line 179: I Believe it should work in the competition, but I think to be on the safe side this should be triggered by autonomous or teleop triggers.

Also, we want to transition to brake mode the moment autonomous or teleop start, not only after 10 seconds. But for the coast mode we do want to wait for some time after the robot has been disabled.

src/main/kotlin/frc/robot/RobotContainer.kt
```diff
@@ -177,6 +168,20 @@ object RobotContainer {
         }
     }
 
+    val disableTrigger: Trigger =
```
katzuv, line 171: private

src/main/kotlin/frc/robot/subsystems/drive/ModuleIOTalonFX.java
```diff
@@ -35,8 +33,11 @@
 import edu.wpi.first.units.measure.AngularVelocity;
 import edu.wpi.first.units.measure.Current;
 import edu.wpi.first.units.measure.Voltage;
+
 import java.util.Queue;
 
+import static frc.robot.lib.PhoenixUtil.tryUntilOk;
```
katzuv, line 39: It's great that you imported this! 
But I don't see you using it. I think it will be good to use this for configuring the neutral mode

katzuv (changes requested):

src/main/kotlin/frc/robot/subsystems/drive/ModuleIOTalonFX.java
yakla: i don't know what is this library then what do you mean by it will be a good to use this for configuring the neutral mode ?

src/main/kotlin/frc/robot/RobotContainer.kt
yakla: ffa0d1d359c63fb53640ba32485a1e64af99c406

src/main/kotlin/frc/robot/RobotContainer.kt
yakla: ffa0d1d359c63fb53640ba32485a1e64af99c406

src/main/kotlin/frc/robot/RobotContainer.kt
yakla: 6f02a79b8524651c6686dc7bd74e6b07404a54c3

src/main/kotlin/frc/robot/RobotContainer.kt
yakla: 6f02a79b8524651c6686dc7bd74e6b07404a54c3

src/main/kotlin/frc/robot/subsystems/drive/ModuleIOTalonFX.java
katzuv: What this function basically does is try to send the motor controller the requested configuration, but sometimes this doesn't work. So this function tries to set the configuration a few times until it works. 
Search for other usages in the project, it will give you an idea how it works.

src/main/kotlin/frc/robot/RobotContainer.kt
katzuv: Great!

src/main/kotlin/frc/robot/subsystems/drive/ModuleIOTalonFX.java
yakla: Also a don't think i was the one who added it it look like its already been used in this class

PR #206, opened 2025-03-30
Merge to main after DCMP

Emma03L:


katzuv (approved):
I'm not going to review this, too much diff. But, this code helps us qualify to the world championship, so I bet it's really good!

